PR #30, opened 2016-03-09
Add gradle bintray dependency example to README

iamdanfox:
The artefact is not published to jcenter yet, so the `dl.bintray.com` thing is necessary here.

The 'latest version badge' in the header came from https://bintray.com/palantir/releases/docker-compose-rule/view.


hpryce:
ACK

PR #32, opened 2016-03-09
Improve Gradle behavior on Circle CI

markelliot:


markelliot:
@hpryce this makes it so that Gradle percentage-complete reports no longer fill the logs, which should make things generally more readable, and possibly run a bit faster.

PR #34, opened 2016-03-10
Feature/make version code make more sense

joelea:
Rather than stripping out the `v` prefix, try again publishing with a flat version number and hope we've managed to get circle to publish with it now.


felixdesouza:
ACK

PR #36, opened 2016-03-11
Feature/explicit service concept

hpryce:
This takes the idea of a service and makes it a real named thing.

Motivations:
- In our project we have a lot of duplication between tests defining healthcheck/waits for common services
- In attempting to solve the above port information became even more scattered all over the place
- While containers are what Docker deals with our ETE tests are using Docker to set up services and it is good to be able to express this in real terms

Concerns:
- To really avoid duplicating ports I needed to make yaml to cover the many cases where the simple setup is sufficient in our uses of this. I didn't want to dump Jackson on the classpath so this is a bit gross.
- The nice thing of having an external file that defines everything is you can run `docker-compose up` on it and replicate the test environment. This loses that but the tests could create a concat'd version of the docker-compose.yml that is used and dump it somewhere so people can play with it still.
- This is only really useful if you have the modern networking stuff in Docker 1.10. Probably about time we dump support for older versions rather than letting us hold it back.


hpryce:
I put @joelea on this but I think @jamesneale and @felixdesouza are likely interested as well.

joelea:
I am super suspicious of the inline compose file stuff - but I'm not going to say it doesn't make sense for your project.

However, I'm really not sure about embedding services in the composition. Would it make more sense to make something like a ServicesRule which builds a composition underneath?

src/main/java/com/palantir/docker/compose/service/ComposePortMappings.java
```diff
@@ -0,0 +1,43 @@
...
+    public ComposePortMappings() {
+        ports = Collections.emptyList();
+    }
+
+    private ComposePortMappings(List<ComposePortDefinition> ports) {
+        this.ports = ports;
+    }
+
+    public ComposePortMappings withPort(ComposePortDefinition composePortDefinition) {
+        ImmutableList<ComposePortDefinition> newPorts = ImmutableList.<ComposePortDefinition>builder()
```
felixdesouza: Nit: Abstract type of `List` please

src/test/java/com/palantir/docker/compose/DockerCompositionIntegrationTest.java
```diff
@@ -28,10 +29,17 @@
...
+    private final DockerService postgresFromComposeFileSnippet = DockerService.fromDockerCompositionFile("src/test/resources/postgres-service.yaml");
+    private final DockerService postgresDefinedInline = DockerService.fromImage("kiasaki/alpine-postgres", "inlinePostgres")
+                                                                     .withPortMapping(5432)
+                                                                     .build();
+
     @Rule
     public DockerComposition composition = DockerComposition.of("src/test/resources/docker-compose.yaml")
                                                             .waitingForService("db", toHaveAllPortsOpen())
                                                             .waitingForService("db2", toHaveAllPortsOpen())
+                                                            .withService(postgresFromComposeFileSnippet)
```
felixdesouza: Looking at how `DockerComposition` is now going to be used, it seems a bit weird that you have a composition `of` some yaml file, which could be whatever you want, and then you have the concept of a service, which is also from a yaml file.

IMO it would make more sense if the file parameter in `of` was gotten rid of, and instead it would consist of a varArgs list of `DockerService`'s.

src/main/java/com/palantir/docker/compose/execution/DockerCompose.java
```diff
@@ -54,6 +55,11 @@ public DockerCompose(DockerComposeExecutable rawExecutable, DockerMachine docker
         this.dockerMachine = dockerMachine;
     }
 
+    public DockerCompose withAdditionalComposeFile(File composeFile) {
```
felixdesouza: I'm not sure this is the right place to put it?

Surely by the time that you have created your `DockerComposition` you would already know the files that you need. It feels like this class will then turn into something that's somewhat a builder but isn't. 

I think action I'm suggesting is: if you really have to, make it a builder, but on the contrary, I don't think it needs to live here since `DockerCompose` isn't public API essentially, and everything should be known by the time a `DockerComposition` is built using its builder.

src/main/java/com/palantir/docker/compose/DockerCompositionBuilder.java
```diff
@@ -17,37 +17,36 @@
...
 
     public DockerCompositionBuilder waitingForService(String serviceName, HealthCheck check) {
-        return waitingForService(serviceName, check, DEFAULT_TIMEOUT);
+        services.add(DockerService.externallyDefined().withHealthCheck(serviceName, check));
+        return this;
     }
 
     public DockerCompositionBuilder waitingForService(String serviceName, HealthCheck check, Duration timeout) {
-        serviceWaits.add(new ServiceWait(containers.get(serviceName), check, timeout));
+        services.add(DockerService.externallyDefined().withTimeout(timeout).withHealthCheck(serviceName, check));
```
felixdesouza: Nit: alignment for consistency sake, here and across the rest of the PR.

src/main/java/com/palantir/docker/compose/configuration/DockerComposeFiles.java
```diff
@@ -38,10 +40,15 @@ public static DockerComposeFiles from(String... dockerComposeFilenames) {
                 .map(File::new)
                 .collect(toList());
         validateAtLeastOneComposeFileSpecified(dockerComposeFiles);
-        validateComposeFilesExist(dockerComposeFiles);
         return new DockerComposeFiles(dockerComposeFiles);
     }
 
+    public DockerComposeFiles withAdditionalFile(File composeFile) {
```
felixdesouza: Similar to the point I mentioned previously, I think all files should be known at some point, so there shouldn't be a need to have a `with...`, and similarly it gets into the whole "I'm a builder but can also represent partial state" if that makes sense.

src/main/java/com/palantir/docker/compose/service/DockerService.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Collections.emptyMap;
+import static org.joda.time.Duration.standardMinutes;
+
+public class DockerService {
```
felixdesouza: Again, I think it would be beneficial to separate out partially constructed objects even if they are validly constructed, simply because you can get into a weird state when actually using it, and then you end up doing horrible horrible things.

felixdesouza:
I think we end falling into a bit of strange territory if we incorporate generating compose files. People will have different needs, and from the surface level of things, it's a very  simple compose file that gets generated. 

A few caveats:
- Currently `docker-compose` has 2 versions of its file format, compose file composition has a requirement for all files in the composition to be of the same version. 
- Somewhat slippery slope, where we then end up writing a docker compose file DSL in java, which would have full support for the numerous options that you can get in a docker-compose file, just to support people's numerous needs. I feel like the moment we have partial support, people would be like: "hey, I would like to use this for my stuff, but your DSL doesn't support this, perhaps I don't want to use it anymore"

felixdesouza:
Regarding concept and motivation, a container in docker seems somewhat synonymous to a service in our use case, in the sense that there isn't more than one service running in a single container, and the health of the container is essentially equivalent to the health of the running service. It feels like unifying that.

The way `DockerService` is defined i.e. it has a `docker-compose.yaml` which could contain a number of services, e.g. a composition. The concept - for me at least - become a bit unclear, because then you will then have a composition of services which could represent another composition. Then an individual service is no longer an individual service, but a collection of services. My mental model then becomes a bit muddled, but that's just me.

hpryce:
I've pushed some code which hopefully refactors to address some of the larger comments.

To response more directly:

Inline service definitions: our hadoop container, for example, is complicated. It runs multiple services as hadoop runs multiple services and those services listen on multiple ports. We need better healthchecks than we have at the moment to make test failures clearer - but the more we add in our own codebase the more that class becomes coupled with the (very simple) docker compose file. By capturing the simple case inline we can have a single class which defines a hadoop service all in one place - so if we change ports or do something else we don't have to go hunting things down all over the place.

Motivation/why not make a wrapper: we have also both pre-setup for some services that I'd like to move out of the docker images to make them less weird (hadoop is a prime example of this - and we use a hacky port signalling method to indicate when presetup is finished inside the container). Making this DockerService class is a set into improving this.

Weirdness with DockerCompositionBuilder/DockerComposeFiles:
- I've killed DockerService.externallyDefined(). This was not a good thing to make.
- No more combining DockerComposeFiles/DockerCompose processes.

Basically, I've compromised. DockerComposition still knows about the very basics of services but the details are contained entirely in the service classes. DockerComposition remains accessible and can be used to configure additional things like saving logs for the entire composition.

src/main/java/com/palantir/docker/compose/service/ComposePortMappings.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.collect.ImmutableList;
+
+import java.util.Collections;
+import java.util.List;
+
+public class ComposePortMappings {
+
+    private final List<ComposePortDefinition> ports;
+
+    public ComposePortMappings() {
```
joelea: Does this ever get used?

src/main/java/com/palantir/docker/compose/service/InlineDockerServiceBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+        return this;
+    }
+
+    public DockerService build() {
+        File dockerComposeFile = buildDockerComposeFile();
+        return DockerService.fromDockerCompositionFile(dockerComposeFile);
+    }
+
+    public DockerService withHealthCheck(HealthCheck healthCheck) {
+        DockerService service = build();
```
joelea: why the random build on this method? Why is there an assumption that this gets called last?

src/main/java/com/palantir/docker/compose/service/ServiceCluster.java
```diff
@@ -0,0 +1,58 @@
...
+    public DockerComposeFiles dockerComposeFiles() {
+        List<File> composeFiles = services.stream()
+                .map(DockerService::dockerComposeFileLocation)
+                .map(Optional::get)
+                .collect(toList());
+        return new DockerComposeFiles(composeFiles);
+    }
+
+    public void addToComposition(DockerCompositionBuilder builder) {
+        services.forEach(service -> service.addWaits(builder));
```
joelea: this seems like it would make more sense reversed and have ServiceCluster implement Iterable.

src/test/java/com/palantir/docker/compose/service/ServiceClusterIntegrationTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.IOException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class ServiceClusterIntegrationTest {
+
+    private final DockerService postgresFromComposeFileSnippet = DockerService.fromDockerCompositionFile("src/test/resources/postgres-service.yaml")
+                                                                              .withHealthCheck("postgres", HealthChecks.toHaveAllPortsOpen());
```
joelea: the `toHaveAllPortsOpen` is designed to be static imported with the `waitForService` method - we should either change `withHealthCheck` or `toHaveAllPortsOpen` - the combination of both is clunky.

src/main/java/com/palantir/docker/compose/service/ServiceCluster.java
joelea: Sorry, now having a better idea of what's going on, just expose a `public Collection<ServiceWaits> getWaits()`?

src/main/java/com/palantir/docker/compose/DockerComposition.java
```diff
@@ -108,4 +110,25 @@ public static DockerCompositionBuilder of(DockerCompose executable) {
         return new DockerCompositionBuilder(executable);
     }
 
+    public static DockerCompositionBuilder fromService(DockerService service) {
```
joelea: I think we might just want to start using a builder except for a few exceptional easy cases. I'd suggest:
`DockerComposition.of(String... paths)` and `DockerComposition.of(DockerService... services)`.

src/main/java/com/palantir/docker/compose/DockerComposition.java
```diff
@@ -108,4 +110,25 @@ public static DockerCompositionBuilder of(DockerCompose executable) {
...
+        return fromServiceCluster(services, DockerMachine.localMachine().build());
+    }
+
+    public static DockerCompositionBuilder fromServiceCluster(ServiceCluster services, DockerMachine dockerMachine) {
+        return fromServiceCluster(new DockerCompose(services.dockerComposeFiles(), dockerMachine), services);
+    }
+
+    public static DockerCompositionBuilder fromServiceCluster(DockerCompose executable, ServiceCluster services) {
+        DockerCompositionBuilder builder = new DockerCompositionBuilder(executable);
+        services.addToComposition(builder);
```
joelea: Yeah, this is the bit that's making me uncomfortable - when Service is just a thing that exposes it's DockerComposeFile and a HealthCheck then this looks like a solid change. 

You can pass around snippets of docker stuff with how you wait for it and inject them in, which I like.

src/test/java/com/palantir/docker/compose/service/ServiceClusterIntegrationTest.java
```diff
@@ -0,0 +1,41 @@
...
+    private final DockerService postgresDefinedInline = DockerService.fromImage("kiasaki/alpine-postgres", "inlinePostgres")
+                                                                     .withPortMapping(5432)
+                                                                     .withHealthCheck(HealthChecks.toHaveAllPortsOpen());
+
+    private final ServiceCluster services = ServiceCluster.of(postgresFromComposeFileSnippet, postgresDefinedInline);
+
+    @Rule
+    public DockerComposition composition = DockerComposition.fromServiceCluster(services)
```
joelea: I should've read the tests first. Now I know what's going on.

I think this should be a more representative example of how we want people to use this, namely:

`DockerComposition.of(fileBackedPostgres, inlinePostgres)`

src/test/java/com/palantir/docker/compose/service/ServiceClusterIntegrationTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class ServiceClusterIntegrationTest {
+
+    private final DockerService postgresFromComposeFileSnippet = DockerService.fromDockerCompositionFile("src/test/resources/postgres-service.yaml")
+                                                                              .withHealthCheck("postgres", HealthChecks.toHaveAllPortsOpen());
+    private final DockerService postgresDefinedInline = DockerService.fromImage("kiasaki/alpine-postgres", "inlinePostgres")
+                                                                     .withPortMapping(5432)
+                                                                     .withHealthCheck(HealthChecks.toHaveAllPortsOpen());
+
+    private final ServiceCluster services = ServiceCluster.of(postgresFromComposeFileSnippet, postgresDefinedInline);
```
joelea: I'm not sure about this, but is there really a difference between ServiceCluster and Service? We could just define a more general thing that bundles a list of files with a list of waits and them smash them together to get larger compositions.

joelea:
This looks pretty exciting now. I can imagine teams being able to publish a bundle which defines health checks and the services and do something like:

`DockerComposition.of(defaultPict(), defaultFoundry(), pictConfiguredMagritte())`

which would be cool.

I just we should be as explicit as possible that the services are just a value type that present a docker-compose file and a healthcheck that is then unpacked into a List<docker-compose-files> and a List<Healthcheck>

palantir-mphi:
This functionality would be pretty useful for some of the Skylab docker-compose stuff I'm working on for Salt & Lime.

joelea:
@hpryce do we have a plan for making progress on this?

joelea:
I'm going to close this until someone does some work on it.

PR #41, opened 2016-04-01
Fix underscore bug in container names

joelea:
See test for bug.

### DIFF
--- a/src/test/java/com/palantir/docker/compose/connection/ContainerNamesTest.java
+++ b/src/test/java/com/palantir/docker/compose/connection/ContainerNamesTest.java
@@ -37,7 +37,7 @@ public class ContainerNamesTest {
     }
 
     @Test
-    public void containers_with_underscores_in_their_naem_are_allowed() {
+    public void containers_with_underscores_in_their_name_are_allowed() {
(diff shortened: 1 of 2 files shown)


felixdesouza:
Although this works for the default cases, if you decide to set your container name for say hostname/linking reasons or whatever, it will fail to detect. Or is that coming in a separate PR?

joelea:
Yeah, this isn't the longterm fix, just a quick fix for a super dumb bug that exists in the implementation we *wanted*. There's some longer term work to fix the bigger thing.

src/test/java/com/palantir/docker/compose/connection/ContainerNamesTest.java
```diff
@@ -37,6 +37,12 @@ public void ps_output_with_a_single_container_results_in_a_single_container_name
     }
 
     @Test
+    public void containers_with_underscores_in_their_naem_are_allowed() {
```
jamesneale: nit: naem -> name

joelea: Thanks.

PR #42, opened 2016-04-04
Healthchecks should give error messages

CRogers:
Tired of just seeing `Container 'myContainer' failed to pass startup check` rather than *why* it failed the startup check, I've now made it so that healthchecks return an error as well as if it succeeded or failed.

### DIFF
--- a/src/main/java/com/palantir/docker/compose/connection/waiting/ServiceWait.java
+++ b/src/main/java/com/palantir/docker/compose/connection/waiting/ServiceWait.java
@@ -22,7 +22,9 @@ import org.joda.time.Duration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class ServiceWait {
     private static final Logger log = LoggerFactory.getLogger(ServiceWait.class);
@@ -38,13 +40,30 @@ public class ServiceWait {
 
     public void waitTillServiceIsUp() {
         log.debug("Waiting for service '{}'", service);
+        final AtomicReference<Optional<SuccessOrFailure>> lastSuccessOrFailure = new AtomicReference<>(Optional.empty());
         try {
             Awaitility.await()
                     .pollInterval(50, TimeUnit.MILLISECONDS)
                     .atMost(timeout.getMillis(), TimeUnit.MILLISECONDS)
-                    .until(() -> healthCheck.isServiceUp(service));
+                    .until(() -> {
+                        SuccessOrFailure successOrFailure = healthCheck.isServiceUp(service);
+                        lastSuccessOrFailure.set(Optional.of(successOrFailure));
+                        return successOrFailure.succeeded();
+                    });
         } catch (ConditionTimeoutException e) {
-            throw new IllegalStateException("Container '" + service.getContainerName() + "' failed to pass startup check");
+            StringBuilder execptionMessage = new StringBuilder();
+            execptionMessage
+                .append("Container '")
+                .append(service.getContainerName())
+                .append("' failed to pass startup check:\n");
+
+            String errorMessage = lastSuccessOrFailure.get()
+                .flatMap(SuccessOrFailure::failureMessage)
+                .orElse("The healthcheck did not finish before the timeout");
+
+            execptionMessage.append(errorMessage);
+
+            throw new IllegalStateException(execptionMessage.toString());
         }
     }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/com/palantir/docker/compose/connection/waiting/ServiceWait.java
```diff
@@ -38,13 +40,30 @@ public ServiceWait(Container service, HealthCheck healthCheck, Duration timeout)
 
     public void waitTillServiceIsUp() {
         log.debug("Waiting for service '{}'", service);
+        final AtomicReference<Optional<SuccessOrFailure>> lastSuccessOrFailure = new AtomicReference<>(Optional.empty());
         try {
             Awaitility.await()
                     .pollInterval(50, TimeUnit.MILLISECONDS)
```
felixdesouza: Could you extract the `Consumer` into a static method or something? so the await would read like:

```java
...
.until(weHaveSuccess(lastSucessOrFailure))
```

src/main/java/com/palantir/docker/compose/connection/waiting/ServiceWait.java
```diff
@@ -38,13 +40,30 @@ public ServiceWait(Container service, HealthCheck healthCheck, Duration timeout)
...
                     .atMost(timeout.getMillis(), TimeUnit.MILLISECONDS)
-                    .until(() -> healthCheck.isServiceUp(service));
+                    .until(() -> {
+                        SuccessOrFailure successOrFailure = healthCheck.isServiceUp(service);
+                        lastSuccessOrFailure.set(Optional.of(successOrFailure));
+                        return successOrFailure.succeeded();
+                    });
         } catch (ConditionTimeoutException e) {
-            throw new IllegalStateException("Container '" + service.getContainerName() + "' failed to pass startup check");
+            StringBuilder execptionMessage = new StringBuilder();
```
felixdesouza: Similarly build the message elsewhere?

src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureMatchers.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.hamcrest.Description;
+import org.hamcrest.FeatureMatcher;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+
+import static org.hamcrest.Matchers.anything;
+import static org.hamcrest.Matchers.equalTo;
+
+public enum SuccessOrFailureMatchers {;
```
felixdesouza: Extraneous `;`

felixdesouza:
LGTM bar minor nits.

joelea:
Sweet.

src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureMatchers.java
CRogers: This is not extraneous :p It's an enum (so you can never instantiate it).

CRogers: An enum with no variants.

felixdesouza: I did not see that, ha!

joelea: Wait? What? Why? Why isn't this a class? Aaaaaaaaah. 

On the other hand, maybe this is genius.

However, on the 3rd hand, I'm pretty sure it's awful.

CRogers: It is usually given as the advice of "use an enum to make static util classes, so that no one can instantiate it!" (obviously missing that you probably should't have static utils classes). We do actually have one here, so why not? An enum is just a class under the scenes, now you literally cannot ever make a `SuccessOrFailureMatchers`, it's like a better private constructor.

joelea: I agree with all of that, but my counterpoint is:

Isn't every single person who ever reads this class going to spend 30 minutes trying to figure out why a random hanging semicolon isn't causing a compile error? I'd at least bring the semicolon onto a line by itself and make a "// Utility class, so no values" comment.

CRogers: Idk, I thought this was a pretty common pattern (within the first few chapters of effective java, iirc) but I'll change it if it makes the code clearer.

joelea: You and your book learning. Lets leave it, see how it works. If we get more exclamations of confusion then we should probably change it, but I'm happy to be proven wrong on the confusion front.

src/main/java/com/palantir/docker/compose/connection/waiting/ServiceWait.java
CRogers: Nice

CRogers:
All comments fixed.

PR #43, opened 2016-04-04
Update code examples in README

nmiyake:
Update so that examples in the README properly compile against the latest build of the library


joelea:
Thanks.

PR #45, opened 2016-04-05
Feature/allow healthchecks on multiple containers

joelea:
While writing atlas tests I came across a case where I want to healthcheck against a cluster of services as one blob. This lets us do it :)


CRogers:
Only had a brief look, but looks broadly good to me.

PR #46, opened 2016-04-08
Retry if `docker-compose up/ps` fails

CRogers:
Due to using a docker swarm and a load of containers, we're seeing some spurious failures for `up` and `ps` that are resolved by retrying the docker command a second time. This PR adds the option to retry `up` and `ps` commands a number of times before giving up.


src/main/java/com/palantir/docker/compose/DockerCompositionBuilder.java
```diff
@@ -34,15 +35,17 @@
 
 public class DockerCompositionBuilder {
     private static final Duration DEFAULT_TIMEOUT = standardMinutes(2);
+    public static final int DEFAULT_RETRY_ATTEMPTS = 2;
```
CRogers: Begin bikeshedding on how many times to retry by default.

hpryce: 3.14 times

PR #51, opened 2016-04-14
Generate success or failure from lambda

joelea:
@CRogers can you merge this tomorrow if it looks good, would like to use it tomorrow.

### DIFF
--- a/src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureShould.java
+++ b/src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureShould.java
@@ -50,4 +50,30 @@ public class SuccessOrFailureShould {
                 containsString("oh no")
             ))));
     }
+
+    @Test
+    public void succeed_on_a_lambda_that_returns_true() {
+        SuccessOrFailure successFromLambda = SuccessOrFailure.onResultOf(() -> true);
+        assertThat(successFromLambda, is(successful()));
+    }
+
+    @Test
+    public void fail_on_a_lambda_that_throws_an_exception() {
+        SuccessOrFailure failureFromLambda = SuccessOrFailure.onResultOf(() -> {
+            throw new IllegalArgumentException("oh no");
+        });
+
+        assertThat(failureFromLambda,
+                is(failureWithMessage(both(
+                        containsString("IllegalArgumentException")).and(
+                        containsString("oh no")
+                ))));
+    }
+
+    @Test
+    public void fail_on_a_lambda_that_returns_false() {
+        SuccessOrFailure failureFromLambda = SuccessOrFailure.onResultOf(() -> false);
+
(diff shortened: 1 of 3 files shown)


src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureShould.java
```diff
@@ -50,4 +50,30 @@ public void fail_from_an_exception() {
...
+    }
+
+    @Test
+    public void fail_on_a_lambda_that_throws_an_exception() {
+        SuccessOrFailure failureFromLambda = SuccessOrFailure.onResultOf(() -> {
+            throw new IllegalArgumentException("oh no");
+        });
+
+        assertThat(failureFromLambda,
+                is(failureWithMessage(both(
```
felixdesouza: Random aside, is it possible to have a type safe way of knowing the class that was thrown? I'm on my phone and can't see the whole code base.

CRogers: Not if you're just returning a string.

src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureShould.java
```diff
@@ -50,4 +50,30 @@ public void fail_from_an_exception() {
...
+                        containsString("IllegalArgumentException")).and(
+                        containsString("oh no")
+                ))));
+    }
+
+    @Test
+    public void fail_on_a_lambda_that_returns_false() {
+        SuccessOrFailure failureFromLambda = SuccessOrFailure.onResultOf(() -> false);
+
+        assertThat(failureFromLambda, is(failureWithMessage("Attempt to complete healthcheck failed")));
```
CRogers: This kinda worries me, it seems every project I'm involved with runs into this problem of turning booleans in generic "it failed" messages rather than propagating some error message.

CRogers: I guess this won't be used by DCR code, so it's your problem :p

CRogers:
Feel free to merge once you fix the build.

src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureShould.java
joelea: This stuff is always just going to generate error messages, it's the only reason for it existing.

src/test/java/com/palantir/docker/compose/connection/waiting/SuccessOrFailureShould.java
joelea: Yes, that's fair. It would be pretty trivial to add an optional string parameter which is the error message. However, in almost all cases I imagine this being used to send failures when exceptions are caught, rather than having to write this try..catch block yourself everywhere.

PR #52, opened 2016-04-19
add support for docker compose exec

zzrabbits:


src/main/java/com/palantir/docker/compose/execution/DockerComposeVersions.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.docker.compose.execution;
+
+public class DockerComposeVersions {
```
joelea: This would be better as a `DockerComposeVersion` class.

CRogers: Could we use a library like https://github.com/zafarkhaja/jsemver?

src/main/java/com/palantir/docker/compose/execution/RetryingDockerCompose.java
```diff
@@ -64,6 +64,15 @@ public void rm() throws IOException, InterruptedException {
     }
 
     @Override
+    public void exec(DockerComposeExecOption dockerComposeExecOption, String containerName,
```
CRogers: Not sure if we should retry docker execs. I haven't seen them go wrong so much in practice as say creating a container.

src/main/java/com/palantir/docker/compose/execution/DockerComposeVersions.java
joelea: Yes, or use that. That looks like the proper version of the class I was suggesting you write.

src/main/java/com/palantir/docker/compose/execution/RetryingDockerCompose.java
joelea: Agreed - I say we make this a non-retryable one for now and see what happens.

Also, if this gets any more calls, I'm totally making this use annotations and some fancy reflection nonesense.

src/test/java/com/palantir/docker/compose/execution/DockerComposeTest.java
```diff
@@ -124,4 +125,26 @@ public void when_there_is_no_container_found_for_ports_an_i_s_e_is_thrown() thro
         compose.ports("db");
     }
 
+    @Test
+    public void docker_compose_exec_pass_concatenated_arguments_to_executor() throws IOException, InterruptedException {
```
joelea: s/pass/passes

src/test/java/com/palantir/docker/compose/execution/DockerComposeTest.java
```diff
@@ -124,4 +125,26 @@ public void when_there_is_no_container_found_for_ports_an_i_s_e_is_thrown() thro
...
+        when(executedProcess.getInputStream()).thenReturn(toInputStream("docker-compose version 1.7.0rc1, build 1ad8866"));
+        compose.exec(new DockerComposeExecOption(new String[] {"-d"}), "container_1",
+                     new DockerComposeExecArgument(new String[] {"ls"}));
+        verify(executor).execute("-v");
+        verify(executor).execute("exec", "-d", "container_1", "ls");
+    }
+
+    @Test
+    public void docker_compose_exec_fail_if_docker_compose_version_is_prior_1_7() throws IOException, InterruptedException {
```
joelea: s/fail/fails

src/test/java/com/palantir/docker/compose/execution/RetryingDockerComposeShould.java
```diff
@@ -57,6 +57,17 @@ public void calls_up_on_the_underlying_docker_compose() throws IOException, Inte
     }
 
     @Test
+    public void calls_exec_on_the_underlying_docker_compose() throws IOException, InterruptedException {
+        DockerComposeExecOption options = new DockerComposeExecOption(new String[] {"-d", "-i"});
+        String container = "just a name";
```
joelea: make this a constant, will make the tests less cluttered.

src/test/java/com/palantir/docker/compose/execution/RetryingDockerComposeShould.java
```diff
@@ -57,6 +57,17 @@ public void calls_up_on_the_underlying_docker_compose() throws IOException, Inte
...
+        DockerComposeExecArgument arguments = new DockerComposeExecArgument(new String[] {"ls", "-la"});
```
joelea: If you make DockerComposeExecArgument and Options take a var args rather than an array then this will be a lot nicer.

joelea: Would also use static factory methods, possibly (with static imports) something like:

dockerCompose.exec(options("-d", "-i"), "some container", arguments("ls -a"))

src/test/java/com/palantir/docker/compose/execution/DockerComposeTest.java
```diff
@@ -124,4 +125,26 @@ public void when_there_is_no_container_found_for_ports_an_i_s_e_is_thrown() thro
...
+
+        try {
+            when(executedProcess.getInputStream()).thenReturn(toInputStream("docker-compose version 1.5.6, build 1ad8866"));
```
joelea: You want to use the ExpectedException rule for this, or at least @Test(expected=IllegalStateException.class)

src/test/java/com/palantir/docker/compose/execution/DockerComposeTest.java
```diff
@@ -124,4 +125,26 @@ public void when_there_is_no_container_found_for_ports_an_i_s_e_is_thrown() thro
         compose.ports("db");
     }
 
+    @Test
+    public void docker_compose_exec_pass_concatenated_arguments_to_executor() throws IOException, InterruptedException {
+        when(executedProcess.getInputStream()).thenReturn(toInputStream("docker-compose version 1.7.0rc1, build 1ad8866"));
+        compose.exec(new DockerComposeExecOption(new String[] {"-d"}), "container_1",
+                     new DockerComposeExecArgument(new String[] {"ls"}));
+        verify(executor).execute("-v");
```
joelea: This check doesn't belong in this test, I think it's basically covered by the next test.

src/main/java/com/palantir/docker/compose/execution/DockerComposeExecArgument.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.docker.compose.execution;
+
+import java.util.Arrays;
+
+public class DockerComposeExecArgument {
```
joelea: Thanks for making proper types for these!

src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -83,6 +83,36 @@ public void rm() throws IOException, InterruptedException {
...
+        executeDockerComposeCommand(throwingOnError(), fullArgs);
+    }
+
+    private void verifyDockerComposeVersion() throws IOException, InterruptedException {
+        String versionOutput = executeDockerComposeCommand(throwingOnError(), "-v");
+        String version = DockerComposeVersions.parseFromDockerComposeVersion(versionOutput);
+        validState(DockerComposeVersions.versionCompare(version, "1.7") >= 0, "You need at least docker-compose 1.7 to run docker-compose exec");
+    }
+
+    private String[] constructFullDockerComposeExecArguments(DockerComposeExecOption dockerComposeExecOption,
```
joelea: Make the 2 wrapper types return List<String> and then just do:

ImmutableList.builder()
    .add("exec")
    .addAll(options.asList())
    .add(containerName)
    .addAll(arguments.asList())
    .build()

Messing around with arrays is normally a recipe for pain and disaster.

src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -83,6 +83,36 @@ public void rm() throws IOException, InterruptedException {
     }
 
     @Override
+    public void exec(DockerComposeExecOption dockerComposeExecOption, String containerName,
+            DockerComposeExecArgument dockerComposeExecArgument) throws IOException, InterruptedException {
+        verifyDockerComposeVersion();
```
joelea: This is a misleading method name. The better version will depend on what our implementation of Version looks like, but it will be something like:

verifyDockerComposeVersionAtLeast(Version.of(1.7.0))

src/main/java/com/palantir/docker/compose/execution/RetryingDockerCompose.java
hpryce: I win.

iamdanfox:
This is pretty cool feature to have, thanks for writing it Colin!  In lieu of any docs, is it worth mentioning this extra way of interacting with containers on the README - perhaps under the [Accessing services in containers][1] section?

[1]: https://github.com/palantir/docker-compose-rule#accessing-services-in-containers-from-outside-a-container

src/main/java/com/palantir/docker/compose/DockerComposition.java
```diff
@@ -110,9 +110,9 @@ public static DockerCompositionBuilder of(DockerCompose executable) {
         return new DockerCompositionBuilder(executable);
     }
 
-    public void exec(DockerComposeExecOption option, String containerName, DockerComposeExecArgument argument)
+    public void exec(DockerComposeExecOption dockerComposeExecOption, String containerName, DockerComposeExecArgument dockerComposeExecArgument)
             throws IOException, InterruptedException {
-        dockerCompose.exec(option, containerName, argument);
```
joelea: I think the old version was probably better, but they do all need plurals, we're passing multiple options an arguments bundled into a single argument.

src/main/java/com/palantir/docker/compose/execution/DockerComposeExecOption.java
```diff
@@ -16,15 +16,20 @@
 package com.palantir.docker.compose.execution;
```
joelea: Sorry, should've mentioned this yesterday - these should probably be @Immutables classes so someone doesn't spend 3 hours trying to figure out a weird bug because these don't implement equals properly. I'm happy to merge this then have a follow up PR/do it myself.

src/test/java/com/palantir/docker/compose/execution/RetryingDockerComposeShould.java
```diff
@@ -57,17 +56,6 @@ public void calls_up_on_the_underlying_docker_compose() throws IOException, Inte
     }
 
     @Test
-    public void calls_exec_on_the_underlying_docker_compose() throws IOException, InterruptedException {
-        DockerComposeExecOption options = new DockerComposeExecOption(new String[] {"-d", "-i"});
-        String container = "just a name";
-        DockerComposeExecArgument arguments = new DockerComposeExecArgument(new String[] {"ls", "-la"});
-        retryingDockerCompose.exec(options, container, arguments);
-        verifyRetryerWasUsed();
-        verify(dockerCompose).exec(options, container, arguments);
```
joelea: Sorry, this test was valuable, just the verification of the retryer.

zzrabbits:
I fixed all the checkstyle manually because my IDEA coding style setting is different from expected.
Do we have a code style setting available to import?

iamdanfox:
@zzrabbits, checkstyle/autoformat pain should be all fixed now that DCR has the latest baseline (#54).  If this doesn't help, can you let me know? Ideally, everyone should just be able to hit 'autoformat' and everything should work.

PR #53, opened 2016-04-21
ProjectName is no longer forced to be random

iamdanfox:
Occasionally, I need to write a test where container startup times are very slow (e.g. hadoop) but the actual test is fast.  I'd like to be able to leave these containers up so that I can iterate on the body of the test quite quickly.

As a prerequisite to this idea, this PR allows users to set a project name, instead of the randomly generated one.

```java
DockerComposition.of(files, machine, "mytest");
```

The default behaviour remains random:

**ProjectName.random()**:
```
CONTAINER ID      IMAGE         COMMAND                  CREATED                  STATUS                  PORTS      NAMES
42e2d6b88bab      app:1.3.0     "/bin/sh -c 'service/"   Less than a second ago   Up Less than a second   8102/tcp,  5adbc76b_app_1
```

**ProjectName.fromString("mytest")**:
```
CONTAINER ID      IMAGE         COMMAND                  CREATED                  STATUS                  PORTS      NAMES
42e2d6b88bab      app:1.3.0     "/bin/sh -c 'service/"   Less than a second ago   Up Less than a second   8102/tcp,  mytest_app_1
```


src/main/java/com/palantir/docker/compose/DockerComposition.java
```diff
@@ -101,7 +102,11 @@ public static DockerCompositionBuilder of(String dockerComposeFile, DockerMachin
     }
 
     public static DockerCompositionBuilder of(DockerComposeFiles dockerComposeFiles, DockerMachine dockerMachine) {
-        return of(new DefaultDockerCompose(dockerComposeFiles, dockerMachine));
+        return of(new DefaultDockerCompose(dockerComposeFiles, dockerMachine, ProjectName.random()));
+    }
+
+    public static DockerCompositionBuilder of(DockerComposeFiles dockerComposeFiles, DockerMachine dockerMachine, String projectName) {
```
CRogers: My first instinct is that `projectName` should live on the builder but you need it here to pass through a docker compose. I really don't like this `of(...).builder().methods()` setup, I wish these `of` overloads had been methods on the builder itself.

src/main/java/com/palantir/docker/compose/configuration/ProjectName.java
```diff
@@ -0,0 +1,65 @@
...
+    @Value.Check
+    protected void validate() {
+        checkState(projectName().trim().length() > 0, "ProjectName must not be blank.");
+
+        checkState(validCharacters(),
+                "ProjectName '%s' not allowed, please use lowercase letters and numbers only.", projectName());
+    }
+
+    // Only allows strings that docker-compose-cli would not modify
+    // https://github.com/docker/compose/blob/85e2fb63b3309280a602f1f76d77d3a82e53b6c2/compose/cli/command.py#L84
```
CRogers: Nice.

src/test/java/com/palantir/docker/compose/configuration/ProjectNameTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.List;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+
+public class ProjectNameTest {
```
CRogers: Nice tests

src/test/java/com/palantir/docker/compose/configuration/ProjectNameTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void construct_compose_command_should_use_project_name_prefix() {
+        List<String> command = ProjectName.random().constructComposeFileCommand();
+
+        assertThat(command, hasSize(2));
+        assertThat(command.get(0), is("--project-name"));
+    }
+
+    @Test
+    public void successive_calls_to_random_should_produce_different_names() {
```
CRogers: Flaky test! I think we can expect this to fail 1/4294967296 times!

iamdanfox: Fight me

src/main/java/com/palantir/docker/compose/DockerComposition.java
iamdanfox: Agreed. One for a big refactoring PR?

PR #55, opened 2016-04-29
Improved builder idea

iamdanfox:
The top level API of a `DockerComposition` has become kinda complicated... there are a bunch of overloaded `of` methods, some of which magically pass things to a new `DefaultDockerCompose` constructor and then other bits of logic on our own hand-written `DockerCompositionBuilder`.

I wanted to write some container re-use logic and ended up refactoring the top level API first (hence the PR).  I use it as follows:

```java
    private static DockerComposeRule docker = DockerComposeRule.builder()
            .saveLogsTo(logsLocation(DemoPathTest.class))
            .files(findResourceFor(DemoPathTest.class))
            .addServices(WEB, SELENIUM)
            .addAllServices(LARGE_MICROSERVICE_STACK)
            .addServices(HADOOP, HDFS_PROXY)
            .projectName(ProjectName.fromString("demopathtest"))
            .build();
```

Where the services above can be conveniently re-used across tests.

This PR maintains the old `DockerComposition` interface, but moves the logic into a new `DockerComposeRule` class.  The idea would be to offer both API points for a while and deprecate one if it makes sense down the line.


src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -0,0 +1,128 @@
...
+            .build();
+    }
+
+    @Value.Default
+    protected DockerCompose dockerCompose() {
+        DockerCompose dockerCompose = new DefaultDockerCompose(executable(), machine());
+        return new RetryingDockerCompose(DockerCompositionBuilder.DEFAULT_RETRY_ATTEMPTS, dockerCompose);
+    }
+
+    @Value.Default
```
CRogers: Why is this `Default`? Do we inject a test double? Otherwise it should be `Derived`.

src/main/java/com/palantir/docker/compose/connection/waiting/MultiServiceHealthCheck.java
```diff
@@ -1,35 +0,0 @@
...
-package com.palantir.docker.compose.connection.waiting;
-
-import static com.google.common.collect.Iterables.getOnlyElement;
-
-import com.google.common.base.Preconditions;
-import com.palantir.docker.compose.connection.Container;
-import java.util.List;
-
-@FunctionalInterface
-public interface MultiServiceHealthCheck {
```
CRogers: I think @joelea was using this class?

src/main/java/com/palantir/docker/compose/DockerService.java
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package com.palantir.docker.compose;
+
+import com.palantir.docker.compose.connection.waiting.SingleServiceHealthCheck;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface DockerService {
```
CRogers: I'm not sure I would call the combination of a service name and healthcheck a `DockerService`? Is this only to make immutables work? There is a sneaky way to write custom methods for your builders.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
iamdanfox: It was a [constructor argument][1] to the old `DockerComposition` class and I wanted to minimise the number of breaking changes in this PR.  If someone was using their own ContainerCache before, they still can.

If you think it's not sensible behaviour, I can definitely remove it as a plugin point?

[1]:https://github.com/palantir/docker-compose-rule/pull/55/files#diff-d1414afe60d59ac1c4f59661e0b22065L47

src/main/java/com/palantir/docker/compose/DockerService.java
iamdanfox: I'm not attached to the name, it's just what we already had.  The aim was to encapsulate the information require to 'wait for a service' in one object.

src/main/java/com/palantir/docker/compose/connection/waiting/MultiServiceHealthCheck.java
iamdanfox: Yep he just explained it is useful for cases where you can't determine a cluster is healthy by looking at each container in turn and need more general logic that can look at a selection of containers.

joelea: Had the conversation on hipchat. Short version is we're going to put this back, this is useful when you can't tell the health of a composition by just looking at services individually.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
joelea: I'm pretty sure there's some test double injection.

joelea:
I think doing a clean overhaul of the top-level builders makes total sense, someone needed to do it :)

I think we still want the individual wait command concept, I'm not sure what adding the service command here is going to give us. Healthchecks seem like a pretty I'm stable, reusable thing, container names don't. However, if your team has a usecase for it then I'm fine with it, but I think this will force most people to create unnecessary indirection by creating a Service then consuming it once, when `waitingFor("db", toBeUp())` would be better.

One option might be to make a service static factory but use the `waitingFor` method name:

`waitingFor(service("db", toRespondToSql()))`

iamdanfox:
@joelea OK so to keep back-compat with both `MultiServiceWait` and `SingleServiceWait`, I've killed off the `DockerService` class.  We now have a `ClusterWait` interface which both `MultiServiceWait` and `SingleServiceWait` both implement:

```java
public interface ClusterWait {
    void waitUntilReady(ContainerAccessor containers);
}
```

This means that the top level `Builder` has exactly the same methods as before:

```java
  private static DockerComposeRule classRule = DockerComposeRule.builder()
            .saveLogsTo(logsLocation(DemoPathTest.class))
            .files(findResourceFor(DemoPathTest.class))
            .waitingForService("selenium", healthCheck)
            .waitingForServices(ImmutableList.of("web1", "web2"), multiCheck)
            .build()
```

Thoughts?

src/main/java/com/palantir/docker/compose/connection/ContainerAccessor.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package com.palantir.docker.compose.connection;
+
+public interface ContainerAccessor {
```
joelea: `Containers` and `containers.get("db")`?

iamdanfox: I almost did this... using a longer method than just `get` means that both the `ContainerCache` and the underlying `DockerCompose` actually conform to this interface, which seemed oddly pleasing.

What about calling it `Containers` but keeping the `container(String name)` method?

joelea: I'm indifferent to the method name. If this is supposed to be a common interface that multiple things implement `Cluster` might be a good name or `ContainerList`.

PR #56, opened 2016-05-03
Skip down/kill/rm when DockerComposeRule#skipShutdown is true.

jkozlowski:


iamdanfox:
:+1:

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -99,10 +104,17 @@ public void before() throws IOException, InterruptedException {
     @Override
     public void after() {
         try {
-            log.debug("Killing docker-compose cluster");
-            dockerCompose().down();
-            dockerCompose().kill();
-            dockerCompose().rm();
+            if (skipShutdown()) {
+                log.error("******************************************************************************************\n"
+                        + "******** Skipping docker-compose down/kill/rm; This should not happen on CI! *************\n"
```
joelea: Can you make it clear that the rule has been configured this way, it's not some weird docker error.

README.md
```diff
@@ -149,6 +149,25 @@ public class DockerCompositionTest {
...
 
+Skipping shutdown
+-----------------
+
+To skip shutdown of containers after tests are finished executing:
+
+```java
+public class DockerCompositionTest {
+
+    private static final boolean SKIP_SHUTDOWN = true;
```
joelea: This is probably less clear than just using the boolean in this case.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
iamdanfox: I actually don't like the `This should not happen on CI!` message.

jkozlowski: How about (with error still):

> docker-compose-rule has been configured to skip docker-compose shutdown: 
> This means the containers will be left running after tests finish executing. 
> If you see this message when running on CI it means you are potentially abandoning
> long running processes.

README.md
jkozlowski: Cool, will update to just call skipShutdown(true), is that what you mean?

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
jkozlowski: Updated the PR.

README.md
jkozlowski: Updated PR.

joelea:
LGTM, will merge and release when builds pass.

jkozlowski:
Sweet, thanks!

PR #57, opened 2016-05-06
Add ip address to Cluster

joelea:


src/main/java/com/palantir/docker/compose/connection/Cluster.java
```diff
@@ -4,8 +4,16 @@
...
 package com.palantir.docker.compose.connection;
 
-public interface Cluster {
+import org.immutables.value.Value;
 
-    Container container(String name);
+@Value.Immutable
+public abstract class Cluster {
+
+    public abstract String ip();
```
iamdanfox: How would this affect your swam usage @CRogers?

joelea:
Consolidated into #59

PR #59, opened 2016-05-06
Feature/use host networked ports

joelea:


src/main/java/com/palantir/docker/compose/connection/waiting/HealthCheck.java
```diff
@@ -13,12 +13,9 @@
...
  */
-
 package com.palantir.docker.compose.connection.waiting;
 
-import com.palantir.docker.compose.connection.Container;
-
 @FunctionalInterface
-public interface SingleServiceHealthCheck {
-    SuccessOrFailure isServiceUp(Container container);
+public interface HealthCheck<T> {
```
iamdanfox: This is nice

iamdanfox:
ACK pending a green build

iamdanfox:
I didn't catch this in code review, but changing `ClusterWait` from an interface to a class broke our integration in the project I'm working on.

Will submit a PR with a representative usage test to avoid this kind of back compat break again!

joelea:
Sorry, this is probably a good case for starting to do sensible release notes/some sort of semver. I was aware this would not be back-compat. I think we're still in the state where we need to prioritise getting the right abstractions rather than maintaining them.

PR #67, opened 2016-05-25
Remove custom Gradle IDEA JDK logic

### LINKED ISSUE #66: Gradle Idea task assumes specific IDEA JDK naming scheme
Currently, this project's Gradle IDEA task assumes as specific naming scheme for the IntelliJ JDK/JRE -- specifically, "JavaSE-$version".

However, this isn't really standard, and makes it so that the out-of-the-box configuration doesn't work in many environments. Is there a reason that this is enforced? Omitting this custom logic should cause IntelliJ to just pick the correct JDK according to the version specified (verified locally and it seems to work)

nmiyake:
Fixes #66.

### DIFF
--- a/gradle/idea.gradle
+++ b/gradle/idea.gradle
@@ -19,11 +19,6 @@ idea {
         inheritOutputDirs = true
         inheritOutputDirs = true
         iml {
-            withXml {
-                def rootManager = it.asNode().component.find { it.@name == 'NewModuleRootManager' }
-                rootManager.@LANGUAGE_LEVEL = jdkVersion
-                rootManager.find {it.@type == "jdk" }.@jdkName = "JavaSE-$javaVersion"
-            }
             whenMerged { module ->
                 module.dependencies.findAll { it.scope == 'TEST' }*.exported = true
             }


nmiyake:
Previously, task generated following IDEA IML:

```
<component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_8" inherit-compiler-output="true">
...
<orderEntry type="jdk" jdkName="JavaSE-1.8" jdkType="JavaSDK" />
```

With this PR, it generates the following:

```
<component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_8" inherit-compiler-output="true">
<orderEntry type="jdk" jdkName="1.8" jdkType="JavaSDK" />
```

Basically, desire is to allow the JDK name to be handled by the plugin rather than specified. Unless there's a reason that an entire workspace has to specifically use the "JavaSE-" naming scheme, don't think it should be assumed.

CRogers:
Hmm, odd that piece of code is in there. I don't have the JavaSE naming either.

CRogers:
You need to merge in master to make this branch mergable.

felixdesouza:
+1 for this.

nmiyake:
Merged in master

joelea:
Thanks @nmiyake !

PR #69, opened 2016-05-25
Allow DockerType to be specified in localMachine

### LINKED ISSUE #68: Docker-compose does not work on Macs that use Docker daemon
The Docker for Mac Beta (https://blog.docker.com/2016/03/docker-for-mac-windows-beta/) finally allows Docker to run as a daemon on Macs rather than as a docker-machine. This is great, but breaks the current assumptions of docker-compose-rule, which hard-codes the `DockerType` for `LocalMachine` to `REMOTE` on Mac OS systems.

Most flexible near-term fix is to allow the `DockerType` to be specified for `LocalMachine` so that developers can customize as necessary. Long-term, it should probably be possible to just eliminate OS switching entirely and allow users to think in terms of just daemon versus remote independently of platform (maybe something to think about for next major version/breaking change).

nmiyake:
Fixes #68 by providing ability to set LocalMachine
to be explicitly set to DAEMON regardless of platform.

### DIFF
--- a/src/main/java/com/palantir/docker/compose/configuration/DockerType.java
+++ b/src/main/java/com/palantir/docker/compose/configuration/DockerType.java
@@ -15,15 +15,16 @@
  */
 package com.palantir.docker.compose.configuration;
 
+import com.google.common.base.StandardSystemProperty;
+
 public enum DockerType {
 
     DAEMON, REMOTE;
 
-    public static final String OS_NAME = "os.name";
     public static final String MAC_OS = "Mac";
 
-    public static DockerType getLocalDockerType() {
-        if (System.getProperty(OS_NAME, "").startsWith(MAC_OS)) {
+    public static DockerType getDefaultLocalDockerType() {
+        if (StandardSystemProperty.OS_NAME.value().startsWith(MAC_OS)) {
             return REMOTE;
         } else {
             return DAEMON;
--- a/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
+++ b/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
@@ -59,7 +59,11 @@ public class DockerMachine implements DockerConfiguration {
     }
 
     public static LocalBuilder localMachine() {
-        return new LocalBuilder(DockerType.getLocalDockerType(), System.getenv());
+        return localMachine(DockerType.getDefaultLocalDockerType());
+    }
+
+    public static LocalBuilder localMachine(DockerType dockerType) {
+        return new LocalBuilder(dockerType, System.getenv());
     }
 
     public static class LocalBuilder {
--- a/src/test/java/com/palantir/docker/compose/configuration/DockerTypeTest.java
+++ b/src/test/java/com/palantir/docker/compose/configuration/DockerTypeTest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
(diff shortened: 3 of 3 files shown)


joelea:
lgtm, needs a master merge.

nmiyake:
Re-pushed.

Just curious, why does your product enforce strict master merge on branches before committing -- is it just to avoid semantic merge conflicts/build errors? I guess this works fine for relatively low-commit-traffic projects, but for most Elements/DI projects with high-traffic, we will typically merge PRs as long as there aren't explicit merge conflicts even if they're not fully up-to-date with master (otherwise we can get into merge wars where a single commit needs to be updated 3-4 times before it can merge because other stuff goes in). Not specifically advocating for you guys to change this, but just wanted to know the reasoning to make sure our other teams aren't missing anything :)

joelea:
It's a default we've had no reason to change. I'm pretty sure the project also existed before that feature did.

joelea:
Though the fact that this has come up is probably a good reason to fix it.

nmiyake:
What's the feature? I'm not actually talking about squash merging or anything, just merging in a PR as long as it doesn't have a merge conflict (which I'm pretty sure has always existed? At least, pretty sure `deployd` has been doing it for almost a year now). If not, then on the technical side it's just a matter of unchecking "Require branches to be up to date before merging" in "branch protection" for any protected branches and then people just merging in even if PR is not fully up-to-date.

Separately, build failed, but doesn't seem to be related to this issue -- is this a failure you guys have seen  before? Going to re-trigger by closing and re-opening request.

felixdesouza:
Doesn't not checking that they're up-to-date introduce the possibility of
merging in bad builds?

Unless I've misunderstood, someone could merge in something into develop,
that may not code-wise conflict, but could possibly break in functionality
with your proposed changes. Your code isn't then run against latest version
of develop. I don't think it's like stash where it rebases on top of
current state of develop automatically.

nmiyake:
Correct -- as I stated above, this approach does introduce the possibility of semantic merge conflicts. Comes down to a cost/benefit analysis of dev productivity versus wanting to make sure every commit runs through tests on their own before going in. On larger product teams with many commits, we found the overhead was not worth it, and data indicated that there were very few cases of actual semantic merge conflicts (one every few months or something). Auto-rebase tools would certainly help.

Sounds like you guys know the trade-offs and went with this approach, and for a project with low commit volume it probably makes sense. It does have a cost, though -- case in point, this commit was "looks good to merge" 5 hours ago, but because one of my own commits went in first, I had to re-pull, rebase, re-push, and then something in the build system was weird so I'm still dealing with it... That's concretely 30 minutes to an hour that this policy cost. If I have multiple changes I want to commit in the future, it also incentives me to have a single PR with multiple commits, which I find less conceptually clean/less desirable, but may do just to get out of the situation where I know I'll have to rebase multiple times to get things in.

nmiyake:
Definitely fairly minor on the whole, though. Also, ready for review/merge again :)

felixdesouza:
Oops didn't see your previous, previous comment. At least for me it wasn't
a conscious decision.

But yeah, I hear you on the merge wars ha. Would be awesome if builds could
be retriggered automatically.

nmiyake:
Thanks!

@joelea @hpryce , any chance we can get a minor release that includes this change? This change is needed to run DockerCompose tests locally on Macs running Docker for Mac Beta, so would be great to have ASAP to unblock that workflow.

hpryce:
@nmiyake shipped 0.13.0

PR #71, opened 2016-05-26
Remove volumes as part of the rm command

### LINKED ISSUE #70: Add option to remove volumes on shutdown
I just noticed my disk had been gobbled up by multiple dangling volumes, from running tests using this rule.  An option (maybe even defaulting to `true`) to remove volumes on shutdown would be great.

joelea:
Fixes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ public class DockerCompositionTest {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


hpryce:
ACK

PR #73, opened 2016-05-30
Add option to remove containers with conflicting names on startup

### LINKED ISSUE #64: Force purge of existing containers with conflicting name.
After an unclean shutdown, running containers will remain, and hinder a repeated test with the error:
``` "Conflict. The name "/docker-contianer-name" is already in use by container 04a0cb1433e1c325ac43a52d0bfa6b0b2ae6ba6665753d50dab7f235a6768645. You have to remove (or rename) that container to be able to reuse that name.```

Please add an option to the builder interface to catch that, and automatically delete the existing container and retry starting it. I can do a PR if you want.

### PR DESCRIPTION
**Pull request by nmiyake**
Fixes #64

### CODE REVIEW

**Comment by nmiyake** [comment]
@joelea @hpryce

Passing and ready for review.

Overview of main functionality change:
* Add `removeConflictingContainersOnStartup` boolean option to `DockerComposeRule` (default is `false`)
* In the `before()` method of `DockerComposeRule`, add a catch to the first `dockerCompose().up()` call
  * Parse exception message. If `removeConflictingContainersOnStartup` is true and the failure was due to conflicting containers, remove the containers and try running `dockerCompose().up()` again.
    * Exception in `remove()` is ignored. Required to work around CircleCI issue where `remove()` call will fail and return 1, but container is removed for purposes of conflicts.

Removing containers is a `docker` (as opposed to `docker-compose`) command. Because of this, a new `Docker` command is introduced. Class structure mirrors that of `DockerCompose`. This PR refactors APIs to share common logic/code. This means that it is a back-compat/API break (although all breaks should be fairly minor/fixed by just updating references to new names). I have a sample PR (#77) that implements the same fix in a non-backcompat-breaking way (duplicates logic instead).

One thing that is a bit awkward right now is that the "remove conflicting names" behavior operates independently of retry behavior. That means that, if there are conflicting containers, the initial `dockerCompose().up()` will do all of its retries (2 by default) before moving on to the `rm` logic. Working around this would be a bit of a pain -- because retry logic is abstracted using the decorator pattern, it's hard to special-case it. I could create a separate `RemoveConflictingNamesRetryingDockerCompose` class that handles both behaviors and return that depending on the configuration, but not sure that it's much cleaner -- open to feedback on that.

File: circle.yml [hunk 1]
```diff
@@ -11,9 +11,10 @@ machine:
 dependencies:
   pre:
     - sudo pip install docker-compose
+    - sudo cp /usr/bin/docker /usr/local/bin/docker
```
hpryce: Just out of interest, why the cp? Could we not execute Docker in /usr/local/bin if present and fall back to /usr/bin if not? It might be useful to add a comment to explain this step to others looking at the circle.yml in future - especially if Circle changes where Docker is installed.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -70,6 +79,11 @@ public DockerComposeExecutable executable() {
     }
 
     @Value.Default
+    public Docker docker() {
+        return new Docker(DockerExecutable.builder().dockerConfiguration(machine()).build());
```
hpryce: There's quite a lot going on in this line which makes it pretty hard to understand. How about making it more like dockerCompose()? This would also address the executable() method now being ambiguous, there could be dockerComposeExecutable() and dockerExecutable() methods.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -94,6 +108,11 @@ protected boolean skipShutdown() {
     }
 
     @Value.Default
+    protected boolean removeConflictingContainersOnStartup() {
+        return false;
```
hpryce: I don't have a strong opinion either way, but is true a better default? As the intention of this is for testing I'm not sure there are many cases where you'd want to fail when conflicting containers exist.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -102,7 +121,20 @@ protected LogCollector logCollector() {
     public void before() throws IOException, InterruptedException {
         log.debug("Starting docker-compose cluster");
         dockerCompose().build();
-        dockerCompose().up();
+
+        try {
+            dockerCompose().up();
+        } catch (DockerExecutionException e) {
+            Set<String> conflictingContainerNames = getConflictingContainerNames(e.getMessage());
```
hpryce: This no longer seems like exceptional behaviour if we're able to handle and recover from it in hopefully the most common situations.

How about making dockerCompose().up() return an object containing the result? The parsing of the exception names could then move into this object which could be unit tested much more easily than having it all in this class.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -102,7 +121,20 @@ protected LogCollector logCollector() {
...
+            if (removeConflictingContainersOnStartup() && !conflictingContainerNames.isEmpty()) {
+                // if rule is configured to remove containers with existing name on startup and conflicting containers
```
hpryce: Rather than a comment this seems like an excellent place for a log line, if somebody might be confused about where their containers have gone on a CI server logging is much more discoverable.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -187,4 +219,23 @@ public String exec(DockerComposeExecOption options, String containerName,
             return addClusterWait(new ClusterWait(clusterHealthCheck, timeout));
         }
     }
+
+    private void removeDockerContainers(Collection<String> containerNames) throws IOException, InterruptedException {
+        Docker docker = docker();
+        try {
+            docker.rm(containerNames.toArray(new String[containerNames.size()]));
```
hpryce: I would strongly recommend overloading the rm method with both the varargs and a collections based method, having to deal with arrays when interacting with the public interface of another class tends to lead to arrays spreading throughout the codebase in my experience.

hpryce:
I'm fine with breaking backcompat (we're still pre-1.0 deliberately because we thought these sort of changes would be necessary once this was used more widely) so I'm going to focus on this PR rather than the other.

I also like trying to ensure there is no duplication between common requirements of running external executables (error handling, awaiting execution) while still handling the differences between the syntax and behaviour of docker-compose and docker executables. However I found the generics here to be very hard to follow, particularly in things like SynchronousDockerExecutable where the class has no behaviour of its own so you have to search up the inheritance hierarchy to discover what is actually happening. There is also duplication between DefaultDockerCompose/Docker and the DockerComposeExecutable/DockerExecutable classes as both now end up knowing/declaring the names of the underlying commands.

I think there are a couple of fairly small refactors that allows the use of generics to be removed entirely and instead polymorphism is used to provide the different behaviour needed for the two different executables. I'm going to try and comment on the classes directly to explain but please do ask if this seems confusing/not workable - if you were in London I'd sit down and pair on this which obviously isn't an option but I'm happy to try a vtc or screen share if you think it would help.

src/main/java/com/palantir/docker/compose/execution/AbstractCommand.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+
+public abstract class AbstractCommand<T extends Executable> {
+
+    private final String commandName;
+    private final AbstractSynchronousExecutable<T> executable;
+
+    public AbstractCommand(String commandName, AbstractSynchronousExecutable<T> executable) {
```
hpryce: Having the command name here results in duplication between the Executable classes (which know the name of the command they are executing as a requirement of performing their responsibility) and the children of this class. Instead, how about adding another method to the executable interface that provides this information for the purpose of constructing error messages.

src/main/java/com/palantir/docker/compose/execution/AbstractCommand.java
```diff
@@ -0,0 +1,56 @@
...
+        this.commandName = commandName;
+        this.executable = executable;
+    }
+
+    protected final String executeCommand(ErrorHandler errorHandler, String... commands)
```
hpryce: Nit: commands should probably be called arguments.

The way this is used by extending this class and then calling executeCommand is kind of weird - there is no abstract method that children are providing that enables this method to perform the action. So how about using composition in DefaultDockerCompose/Docker? Each of the methods in there would go from executeCommand(errorHandler, commands) to command.execute(errorHandler, commands) which is just as easy to understand when reading those classes but instead this class becomes possible to unit test by itself to ensure the behaviour it has is correct.

src/main/java/com/palantir/docker/compose/execution/AbstractSynchronousExecutable.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.concurrent.TimeoutException;
+import java.util.function.Consumer;
+
+public abstract class AbstractSynchronousExecutable<T extends Executable> {
+    public static final int HOURS_TO_WAIT_FOR_STD_OUT_TO_CLOSE = 12;
+    public static final int MINUTES_TO_WAIT_AFTER_STD_OUT_CLOSES = 1;
+    private final T executable;
+    private final Consumer<String> logConsumer;
+
+    public AbstractSynchronousExecutable(T executable, Consumer<String> logConsumer) {
```
hpryce: Both the generics and the abstract nature of this class are not used, so it seems like additional complexity that isn't required. If the generics here disappear then so do the generics on AbstractCommand which further simplifies things. Making it a concrete class should also enable unit tests to be written to ensure this does actually wait for the process to finish.

hpryce: Continuing my suggestion above to provide the command name from the Executable this would need a pass through method to the underlying Executable to keep AbstractCommand nice and simple. I thought about whether this should implement Executable but I don't think that is the right thing, consumers of this require the synchronous behaviour that isn't guaranteed by the Executable interface and therefore this is it's own type.

src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -56,50 +53,51 @@ public DefaultDockerCompose(DockerComposeFiles dockerComposeFiles, DockerMachine
     }
 
     public DefaultDockerCompose(DockerComposeExecutable rawExecutable, DockerMachine dockerMachine) {
+        super("docker-compose", new SynchronousDockerComposeExecutable(rawExecutable, log::debug));
```
hpryce: My intention with the suggestions above is that this would now become something like:

    public DefaultDockerCompose(DockerComposeExecutable rawExecutable, DockerMachine dockerMachine) {
        this.command = new Command(rawExecutable, log::debug);
        ...
    }

    public void build() {
        command.execute(throwingOnError(), "build");
    }

    ...

Doing this means this class and Docker would no longer share a common parent which matches the actual methods and behaviour on the classes - they both execute a command underneath and need to handle errors etc but that is abstracted away from the consumer of this class. This would avoid somebody using AbstractCommand as a type and then ending up with an object that must be cast to something else before any method can be invoked which is a pattern that I would like to avoid due to the difficultly that arises when adding new polymorphic types that the casting no longer accounts for.

src/main/java/com/palantir/docker/compose/execution/SynchronousDockerComposeExecutable.java
```diff
@@ -15,61 +15,13 @@
...
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 
-public class SynchronousDockerComposeExecutable {
-    public static final int HOURS_TO_WAIT_FOR_STD_OUT_TO_CLOSE = 12;
-    public static final int MINUTES_TO_WAIT_AFTER_STD_OUT_CLOSES = 1;
-    private final DockerComposeExecutable dockerComposeExecutable;
-    private final Consumer<String> logConsumer;
+public class SynchronousDockerComposeExecutable extends AbstractSynchronousExecutable<DockerComposeExecutable> {
```
hpryce: With the changes I suggested above to remove generics I think this class is no longer needed which makes things a lot simpler and easier for somebody new to the codebase.

src/test/java/com/palantir/docker/compose/DockerComposeRuleRestartContainersIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+                DockerExecutable.builder().dockerConfiguration(DockerMachine.localMachine().build()).build());
+        try {
+            docker.rm("/test-1.container.name", "/test-2.container.name");
+        } catch (DockerExecutionException e) {
+            log.debug("docker rm failed in cleanup, but continuing", e);
+        }
+    }
+
+    @Test
+    public void testDockerComposeRuleFailsWithExistingContainers() throws IOException, InterruptedException {
```
hpryce: Nit: We've been using underscores rather than camel cases in our tests names to aid readability.

src/test/java/com/palantir/docker/compose/execution/DockerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.mockito.Matchers.anyVararg;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.io.IOException;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DockerTest {
```
hpryce: In general we've been trying to use the convention that unit tests classes are named "FooShould" with test methods that read as sentences, e.g. DockerShould call_rm_with_the_force_flag.

Sorry that things are not 100% consistent in the codebase on this front yet, we've been refactoring as we touch things to make them conform and we perhaps need a more concerted effort to provide uniformity.

nmiyake:
Thanks for the feedback! Some of the things you pointed out were due to me wanting to keep the footprint of the change set small (and not wanting to put in too much work before vetting that general approach was OK with core team). Will update the change set based on your feedback and re-submit. Thanks!

circle.yml
nmiyake: Main reason was to keep code symmetric -- `DockerComposeExecutable` had its locations as `System.getenv("DOCKER_COMPOSE_LOCATION")` and `/usr/local/bin/docker-compose`, and it works because the "pip install" installs the executable there. Agree this is a bit awkward, though -- I'll just add `/user/bin/<exec>` as a third location for both in the updated PR and remove the extra line from circle.yml

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
nmiyake: Yup, makes sense -- done

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
nmiyake: I went back and forth on this, and played around with returning a result from `dockerCompose.up()`. However, I didn't really like the result -- basically, you would end up in a place where `dockerCompose.up()` could return a value, and the content of the value would need to be inspected to determine whether or not the operation was successful. This just didn't feel right to me, especially if the rest of the contracts weren't being altered.

With the current structure, I think that, if `dockerCompose.up()` returns normally, it should mean that the operation succeeded and that the user can assume that the compose cluster is now "up". Any case besides that should be exceptional.

I've cleaned this up by creating a separate implementation of `DockerComposeRule` that encodes this logic. Its implementation of `up()` delegates to the internal one, catches the exceptional state it supports and then attempts to re-try. This centralizes this logic in a single place, but still keeps the over-all contract -- if the wrapped `up()` returns, it means that the cluster did in fact come up, while if it couldn't recover, it will throw an exception. The logic no longer leaks out to `DockerComposeRule`, and I think it's consistent with the way other things in this library are implemented.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/execution/AbstractCommand.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/execution/AbstractCommand.java
nmiyake: Done for the refactoring. Not sure if I renamed everything to your liking -- feel free to re-comment on newer patch if you see any place you'd like me to update.

src/main/java/com/palantir/docker/compose/execution/AbstractSynchronousExecutable.java
nmiyake: Addressed first note. With the way the logic ended up working out, didn't need to do the passthrough method since only place that `name()` was called had a handle to the raw executable and could call it on that directly.

src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/execution/SynchronousDockerComposeExecutable.java
nmiyake: Done

src/test/java/com/palantir/docker/compose/DockerComposeRuleRestartContainersIntegrationTest.java
nmiyake: Done

src/test/java/com/palantir/docker/compose/execution/DockerTest.java
nmiyake: Done

nmiyake:
@hpryce ready for another round of review. I believe I've addressed most of the major points in your first review and the code looks/feels a lot more consistent with the rest of the codebase now.

src/main/java/com/palantir/docker/compose/execution/ConflictingContainerRemovingDockerCompose.java
```diff
@@ -0,0 +1,98 @@
...
+                super.up();
+                return;
+            } catch (DockerExecutionException e) {
+                Set<String> conflictingContainerNames = getConflictingContainerNames(e.getMessage());
+                if (conflictingContainerNames.isEmpty()) {
+                    // failed due to reason other than conflicting containers, so re-throw
+                    throw e;
+                }
+
+                if (currRetryAttempt == retryAttempts) {
```
hpryce: I actually think in the final error case we also want to remove the containers to avoid a build of crap left around. This would also simplify the logic here a lot as it becomes a for loop.

src/main/java/com/palantir/docker/compose/execution/ConflictingContainerRemovingDockerCompose.java
```diff
@@ -0,0 +1,98 @@
...
+        this.retryAttempts = retryAttempts;
+    }
+
+    @Override
+    public void up() throws IOException, InterruptedException {
+        int currRetryAttempt = 0;
+
+        while (true) {
+            try {
+                super.up();
```
hpryce: This is pretty nitty, but it seems odd to override a method that you can't then actually reimplement yourself. How about making the delegated docker composition protected rather than private?

hpryce:
Happy for this to be merged, points are minor. Needs conflicts resolving though.

nmiyake:
Thanks! Will resolve conflicts and address feedback.

src/main/java/com/palantir/docker/compose/execution/ConflictingContainerRemovingDockerCompose.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/execution/ConflictingContainerRemovingDockerCompose.java
nmiyake: Done. Also added a "protected final DockerCompose getDockerCompose()" method to the superclass that subclasses can use to get the wrapped compose object and act on it directly -- removes the call to `super.*()` methods, which may make intent clearer.

nmiyake:
Rebased and addressed feedback; ready for review

PR #79, opened 2016-06-03
Make default localMachine() choose DockerType based on environment

nmiyake:
Makes it so that default use of DockerCompose can work in both
daemon and remote environments without having to re-write code.
Also refactors DockerType to centralize related logic.


src/main/java/com/palantir/docker/compose/configuration/DockerType.java
```diff
@@ -15,20 +15,41 @@
  */
 package com.palantir.docker.compose.configuration;
 
-import com.google.common.base.StandardSystemProperty;
+import com.google.common.base.Optional;
+import java.util.Map;
 
-public enum DockerType {
+public enum DockerType implements HostIpResolver, EnvironmentValidator {
```
nmiyake: Open to different approach for this part. Basic thing was that each DockerType had resolver and validator behavior associated with it, so wanted to encapsulate in code rather than having the current looser packaging.

Alternatives here are to use a class hierarchy instead (could just have a `DockerType` abstract class that defines `validate` and `resolveIp` and have subclasses provide implementation instead of breaking out into pieces like this) or to not have `DockerType` implement the interfaces but just provide getters that return the interfaces (although I think that just adds a layer of indirection without adding much benefit).

src/main/java/com/palantir/docker/compose/configuration/DaemonEnvironmentValidator.java
```diff
@@ -21,18 +21,26 @@
...
 
-public final class DaemonEnvironmentValidator {
+public final class DaemonEnvironmentValidator implements EnvironmentValidator {
 
     private static final Set<String> ILLEGAL_VARIABLES = ImmutableSet.of(DOCKER_TLS_VERIFY, DOCKER_HOST, DOCKER_CERT_PATH);
+    private static final Supplier<DaemonEnvironmentValidator> SUPPLIER = Suppliers.memoize(
+            () -> new DaemonEnvironmentValidator());
+
+    public static DaemonEnvironmentValidator instance() {
+        return SUPPLIER.get();
```
nmiyake: See comment on `DockerType` for more info. Awkwardness of having a supplier for both validators is because we're basically providing behavior here and wiring it up in the enum (conceptually, we're defining classes in the enum and attaching behavior from these other interfaces using objects). The more idiomatic approach would probably be to use a class hierarchy, and then this implementation would move to a single class.

I would likely prefer the class hierarchy approach, but started with this approach since it more closely mirrors existing code and current codebase seems to bias towards breaking out behavior into separate objects/classes over consolidating in methods.

nmiyake:
@hpryce @joelea 

Would like this approach since it allows tests that use `DockerCompose` to run in Docker Toolbox or Docker for Mac Beta environments without having to change any source code -- as it stands right now, Docker for Mac Beta users have to manually add `.machine(DockerMachine.localMachine(DockerType.DAEMON).build())` to their `DockerCompose` tests for them to properly run locally, and having to continually add/remove this for local testing is tedious

hpryce:
I believe @jkozlowski might be interested in this, he mentioned this at some point.

@felixdesouza / @iamdanfox could you take a look at this? I seem to recall some PRs from you related to this so you're probably better placed to review it than me.

nmiyake:
Rebased and ready for review. @hpryce @felixdesouza @iamdanfox let me know who should be the point person on the review

src/main/java/com/palantir/docker/compose/configuration/EnvironmentValidator.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.docker.compose.configuration;
+
+import java.util.Map;
+
+public interface EnvironmentValidator {
+
+    Map<String, String> validate(Map<String, String> dockerEnvironment);
```
iamdanfox: As someone who hasn't encountered this part of the code before, it wasn't immediately obvious to me what this interface is for.

Perhaps we could use some more descriptive types, method names or javadoc to make this super obvious?  Also I couldn't see why this method actually returns another Map<String,String>... What about:

```java
void validateEnvironmentVariables(Map<String, String> environmentVariables);
```

src/main/java/com/palantir/docker/compose/configuration/RemoteEnvironmentValidator.java
```diff
@@ -23,25 +23,30 @@
...
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class RemoteEnvironmentValidator {
+public final class RemoteEnvironmentValidator implements EnvironmentValidator {
 
     private static final Set<String> SECURE_VARIABLES = ImmutableSet.of(DOCKER_TLS_VERIFY, DOCKER_CERT_PATH);
+    private static final Supplier<RemoteEnvironmentValidator> SUPPLIER = Suppliers.memoize(
```
iamdanfox: Could we get away with just having a regular old static final field here?  How necessary is this memoized supplier?

src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
```diff
@@ -88,22 +98,15 @@ public LocalBuilder withEnvironment(Map<String, String> newEnvironment) {
         }
 
         public DockerMachine build() {
-            HostIpResolver hostIp;
-            if (DockerType.DAEMON == dockerType) {
```
iamdanfox: Nice

src/main/java/com/palantir/docker/compose/configuration/DockerType.java
iamdanfox: This seems pretty nice

src/main/java/com/palantir/docker/compose/configuration/DockerType.java
```diff
@@ -15,20 +15,41 @@
...
+    public static Optional<DockerType> getFirstValidDockerTypeForEnvironment(Map<String, String> environment) {
+        for (DockerType currType : DockerType.values()) {
+            try {
+                currType.validate(environment);
+                return Optional.of(currType);
+            } catch (IllegalStateException e) {
+                // ignore and try next type
+            }
         }
+        return Optional.absent();
```
felixdesouza: Java 8?

iamdanfox:
Looks like a sensible and useful PR, just a couple of comments, otherwise ACK from me

src/main/java/com/palantir/docker/compose/configuration/DockerType.java
iamdanfox: Oooh good spot

src/main/java/com/palantir/docker/compose/configuration/EnvironmentValidator.java
nmiyake: Done -- initially was aiming for a straight refactor, but this change makes sense.

src/main/java/com/palantir/docker/compose/configuration/DockerType.java
nmiyake: Done

src/main/java/com/palantir/docker/compose/configuration/RemoteEnvironmentValidator.java
nmiyake: Done

nmiyake:
Thanks -- addressed PR feedback and ready for review

nmiyake:
@iamdanfox @felixdesouza @hpryce ping for another round of review

iamdanfox:
MERGE MERGE MERGE

iamdanfox:
Tagging someone with merge permissions: @hpryce @joelea @CRogers

nmiyake:
@hpryce @joelea @CRogers -- ping for merge now that this has been CR'd and approved

PR #81, opened 2016-06-23
Don't assume that an SSL exception when health checking a container means it's up.

j-baker:
A service can open a port before it's ready to accept SSL connections.

Not sure how I'd go about testing this one.


iamdanfox:
ACK

CRogers:
Going to merge it anyway - should mostly solve the problem for people only using http, which is most people.

PR #86, opened 2016-07-01
Add option to pull images on startup.

rveguilla:


iamdanfox:
Could you talk me through why you needed this flag?  We already run `docker-compose build` and `docker-compose up`...

rveguilla:
Sure. The problem is that, by default,  `docker-compose build` command does not check if a newer image is available for any previously pulled image tags. This may cause inconsistent test results between different docker hosts running the same tests depending on when each host initially pulled a particular image.

I ran into this problem while trying to build an integration pipeline that tries to test the 'latest'
version of Service A against the 'latest' version of Service B. After fixing a bug in Service B that was causing a test to fail, the test passed on the docker host where the images where built, but not on the docker host used by the CI server, which was still using the old Service B image.

rveguilla:
@iamdanfox any other question?

hpryce:
Hi @rveguilla thanks for the PR - the reasoning behind this seems sound and we'd love to get it merged in.

Would it be possible for you to review our contributing.md: https://github.com/palantir/docker-compose-rule/blob/master/CONTRIBUTING.md
Sorry that this wasn't available when you first made the PR. If you're willing to follow the steps in there we can get this merged and released.

iamdanfox:
Closing and re-opening to trigger the CLAbot

palantirtech:
Thanks for your interest in [palantir/docker-compose-rule](https://github.com/palantir/docker-compose-rule), @rveguilla! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

lcatoni:
Hi @iamdanfox, I am interested in this functionality. Do you have an ETA for this getting merged? Thanks!

iamdanfox:
@lcatoni I'm keen to make DCR's startup behaviour configurable, a few other people have asked for this functionality too!  However, I'd like to do this in a more extensible way than the above PR so that consumers of the library can hack it to do what they want.

Specifically, I'm imagining writing a new `StartupStrategy` interface, somewhat like the recent `ShutdownStrategy`.  We'd ship with a couple of implementations of this (the regular build, up -d behaviour, maybe also a pull, build, up -d), but this would enable users to write their own logic (e.g. start up only a subset of the containers).  @harrybiddle would this cover your use case too?

joelea:
Would a List<StartupAction> be more appropriate? Add an addStartupAction to the builder interface?

harrybiddle:
@iamdanfox - that would do it for me! I am pretty happy with the workaround of bringing up the whole cluster and killing what I don't initially need, so it's not high priority for me, but I'd take it if it was there, just to be a bit cleaner.

joelea:
@iamdanfox are we still happy to merge this? I can go and fix the merge conflicts and get it merged if so.

iamdanfox:
@joelea yeah it seems fine to merge this :) when we get more requests here, I'd like to make this more extensive akin to the `StartupStrategy` thing mentioned above, but it seems like it'll work as a start!

hpryce:
Created https://github.com/palantir/docker-compose-rule/pull/154 in order to get this merged in.

PR #89, opened 2016-07-08
Change maven bintray url from http to https

natgabb:


iamdanfox:
Thanks for this contribution :) good spot!

PR #95, opened 2016-08-05
Add support for calling docker-compose run

SerialVelocity:
Add support for calling docker-compose run. This feature is needed so we can work around the CircleCI issue of not being able to use ```lxc-attach``` or ```docker exec``` on docker 1.10 and higher.

### DIFF
--- a/src/main/java/com/palantir/docker/compose/execution/DockerComposeRunArgument.java
+++ b/src/main/java/com/palantir/docker/compose/execution/DockerComposeRunArgument.java
@@ -22,13 +22,9 @@ import org.immutables.value.Value;
 @Value.Immutable
 public abstract class DockerComposeRunArgument {
     @Value.Parameter
-    protected abstract String[] arguments();
+    public abstract List<String> arguments();
 
     public static DockerComposeRunArgument arguments(String... arguments) {
-        return ImmutableDockerComposeRunArgument.of(arguments);
-    }
-
-    public List<String> asList() {
-        return Arrays.asList(arguments());
+        return ImmutableDockerComposeRunArgument.of(Arrays.asList(arguments));
     }
 }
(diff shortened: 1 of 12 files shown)


iamdanfox:
:+1:

src/main/java/com/palantir/docker/compose/execution/DockerComposeRunArgument.java
```diff
@@ -0,0 +1,34 @@
...
+package com.palantir.docker.compose.execution;
+
+import java.util.Arrays;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class DockerComposeRunArgument {
+    @Value.Parameter
+    protected abstract String[] arguments();
```
felixdesouza: Make this a `List<String>` and do the conversion in the constructor?

felixdesouza: Which also get's rid of the `asList` method.

src/main/java/com/palantir/docker/compose/execution/DockerComposeRunOption.java
```diff
@@ -0,0 +1,34 @@
...
+package com.palantir.docker.compose.execution;
+
+import java.util.Arrays;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class DockerComposeRunOption {
+    @Value.Parameter
+    protected abstract String[] options();
```
felixdesouza: Same as in `DockerComposeRunArgument`

src/main/java/com/palantir/docker/compose/execution/DockerComposeRunArgument.java
SerialVelocity: I'll change DockerComposeExecArgument too since this is just a c&p of that.

PR #98, opened 2016-08-31
Add method to kill a container

hsaraogi:
Stop(Container) sends a SIGTERM followed by a SIGKILL to a container that allows time for graceful shutdown. We want to kill a node i.e. a container to simulate node dying. 

Your PR should contain a few sentences describing what problem your PR solves.


CRogers:
👍 

Rerunning build to see if it was a flake.

PR #101, opened 2016-09-07
Don't have confusing Builder class in DockerComposeRule

### LINKED ISSUE #100: DockerComposeRule.Builder is confusing
Either DCR.Builder should be package private, or should extend ImmutableDCR.Builder and not the other way around.

It's super confusing that you can do

    DockerComposeRule.Builder builder = DockerComposeRule.builder()

and then can't actually call `build()`.

My understanding is that the pattern you're *meant* to use with that Immutables feature is to declare the outer class as AbstractDockerComposeRule.Builder and then you use the Immutables generated version everywhere. The current implementation seems to be a bit worse than just having it extend ImmutableDCR.Builder. The other acceptable solution would be to make DCR.Builder package private, as then you can't use the type DockerComposeRule.Builder in your code.

j-baker:
Fixes #100.

### DIFF
--- a/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
+++ b/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
@@ -165,49 +165,43 @@ public abstract class DockerComposeRule extends ExternalResource {
         return dockerCompose().exec(options, containerName, arguments);
     }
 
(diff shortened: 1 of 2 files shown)


iamdanfox:
@j-baker can you bump this or shall I cherry-pick your commit?

iamdanfox:
Have fixed merge conflicts and opened #110

PR #106, opened 2016-09-30
Fix depluralize() compile warning

tomasulo:
This PR fixes a compile warning concerning `depluralize()`. It is the same issue that was addressed here: https://github.com/palantir/docker-compose-rule/commit/d929d35e936ae6915a2f3ff8924e87fba9a4d888


iamdanfox:
Thanks @tomasulo, always nice to clear out some warnings!

PR #112, opened 2016-10-02
Faster integration test

iamdanfox:
Also includes a tiny bit of nicer logging.

Individual tests are faster now, and the overall build times are about 5m40.


iamdanfox:
@a-k-g CR'd

PR #114, opened 2016-10-03
Add a noOptions method to DockerComposeExecOption

### LINKED ISSUE #113: Overloading issue in DockerComposeExecOption
We never want to provide an option to `docker-compose exec`. However, this makes it a pain to actually construct an object of type `DockerComposeExecOption`, because the type has

`public static DockerComposeExecOption options(String... options)` and `public abstract List<String> options()` which collide and stop code from compiling.

Our code now looks like `DockerComposeExecOption.options(new String[0])`

j-baker:
Non-breaking change to fix an overloading bug.

Fixes #113.

34d64fe made the protected instance method `options()` public, which
makes it clash with the public static method `options(String...)` in the
frequent case you don't want to pass any options.

This PR adds a `noOptions()` method which avoids this kind of bad
behaviour.

I did this because everything else would break at least one person. More
generally, it's probably advisable to try to name static methods
differently to instance methods, or you open yourself up to these
overloading based bugs.

### DIFF
--- a/src/main/java/com/palantir/docker/compose/execution/DockerComposeExecOption.java
+++ b/src/main/java/com/palantir/docker/compose/execution/DockerComposeExecOption.java
@@ -15,9 +15,9 @@
  */
 package com.palantir.docker.compose.execution;
 
+import com.google.common.collect.ImmutableList;
 import java.util.Arrays;
 import java.util.List;
-
 import org.immutables.value.Value;
 
 @Value.Immutable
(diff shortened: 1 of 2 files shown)


iamdanfox:
Checkstyle failure, "Line does not match expected header line" yada yada

(Other than that, looks fine to me)

https://389-50542309-gh.circle-artifacts.com/0/home/ubuntu/docker-compose-rule/build/reports/checkstyle/test.html#f-/home/ubuntu/docker-compose-rule/src/test/java/com/palantir/docker/compose/execution/DockerComposeExecOptionShould.java

PR #115, opened 2016-10-04
Integration test verifies AggressiveShutdownStrategy actually works

iamdanfox:
Turns out if you run `docker-compose rm --force --volumes` with containers it does absolutely nothing.

Instead, we will now do `docker rm -f proj_myservice_1 proj_otherservice_1 proj_another_1`, which actually does what we want.

I've had to add some more things because circle spews btrfs errors.

@felixdesouza


src/main/java/com/palantir/docker/compose/connection/ContainerNames.java
```diff
@@ -16,98 +16,26 @@
...
-            return false;
-        }
-        ContainerNames other = (ContainerNames) obj;
-        return Objects.equals(containerNames, other.containerNames);
-    }
-
-    @Override
-    public String toString() {
-        return "ContainerNames [containerNames=" + containerNames + "]";
+        String psBody = splitOnSeparator[1];
```
felixdesouza: Can you put this in a method? I don't really know what's going on even after reading it.

src/main/java/com/palantir/docker/compose/connection/ContainerNames.java
```diff
@@ -16,98 +16,26 @@
...
 
 import java.util.Arrays;
-import java.util.Iterator;
 import java.util.List;
-import java.util.Objects;
-import java.util.function.Function;
-import java.util.stream.Stream;
 
-public class ContainerNames implements Iterable<String> {
+public class ContainerNames {
```
felixdesouza: This is very sensible.

src/main/java/com/palantir/docker/compose/execution/GracefulShutdownStrategy.java
```diff
@@ -18,11 +19,11 @@
     private static final Logger log = LoggerFactory.getLogger(GracefulShutdownStrategy.class);
 
     @Override
-    public void shutdown(DockerCompose dockerCompose) throws IOException, InterruptedException {
+    public void shutdown(DockerComposeRule rule) throws IOException, InterruptedException {
         log.debug("Killing docker-compose cluster");
-        dockerCompose.down();
-        dockerCompose.kill();
-        dockerCompose.rm();
+        rule.dockerCompose().down();
```
felixdesouza: What's the benefit of passing in the rule?

src/test/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    public void shut_down_multiple_containers_immediately() throws Exception {
+        DockerComposeRule rule = DockerComposeRule.builder()
+                .file("src/test/resources/shutdown-strategy.yaml")
+                .logCollector(new DoNothingLogCollector())
+                .retryAttempts(0)
+                .shutdownStrategy(ShutdownStrategy.AGGRESSIVE)
+                .build();
+
+        assertThat(rule.dockerCompose().ps(), is(TestContainerNames.of()));
```
felixdesouza: Not entirely sure this assertion is useful.

src/test/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+        rule.before();
```
felixdesouza: Might be worth asserting that the containers are non-empty, so we can know that we're actually killing things.

I suppose it's also quite hard to test that things are aggressively removed :S

src/test/java/com/palantir/docker/compose/TestContainerNames.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private TestContainerNames() {}
+
+    public static List<ContainerName> of(String... semanticNames) {
+        return Arrays.stream(semanticNames)
+                .map(TestContainerNames::testContainerName)
+                .collect(toList());
+    }
+
+    private static ImmutableContainerName testContainerName(String testName) {
```
felixdesouza: Do we make use of the immutable implementation?

felixdesouza: Can we just return `ContainerName`?

src/test/java/com/palantir/docker/compose/TestContainerNames.java
```diff
@@ -0,0 +1,31 @@
...
+package com.palantir.docker.compose;
+
+import static java.util.stream.Collectors.toList;
+
+import com.palantir.docker.compose.connection.ContainerName;
+import com.palantir.docker.compose.connection.ImmutableContainerName;
+import java.util.Arrays;
+import java.util.List;
+
+public class TestContainerNames {
```
felixdesouza: Actually not entirely sure this class is useful. In other instances we're making our own instances of `ContainerName`

felixdesouza: Or maybe we can use this is more places, e.g. `ContainerNamesShould`

src/test/java/com/palantir/docker/compose/connection/ContainerNamesShould.java
```diff
@@ -15,49 +15,56 @@
...
+import java.util.List;
 import org.junit.Test;
 
 public class ContainerNamesShould {
 
     @Test
     public void result_in_no_container_names_when_ps_output_is_empty() {
-        ContainerNames names = ContainerNames.parseFromDockerComposePs("\n----\n");
-        assertThat(names, is(new ContainerNames(emptyList())));
+        List<ContainerName> names = ContainerNames.parseFromDockerComposePs("\n----\n");
```
felixdesouza: Do we need all of these tests? Most of the functionality has moved into `ContainerName` so they're probably replicated there.

src/main/java/com/palantir/docker/compose/connection/ContainerNames.java
iamdanfox: Done

src/main/java/com/palantir/docker/compose/execution/GracefulShutdownStrategy.java
iamdanfox: It means that any implementations of the `ShutdownStrategy` interface have maximal functionality available to them

src/test/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
iamdanfox: I actually wanted to keep this because it's a precondition for the last assertion to make sense

src/test/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
iamdanfox: Done

src/test/java/com/palantir/docker/compose/TestContainerNames.java
iamdanfox: done

src/test/java/com/palantir/docker/compose/TestContainerNames.java
iamdanfox: I think this still has merit, because this one method is used in a bunch of different places to create hacky ContainerName. `ContainerNamesShould` actually creates fully legit ones

felixdesouza: Fair

src/test/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
felixdesouza: Can you elaborate?

src/main/java/com/palantir/docker/compose/execution/GracefulShutdownStrategy.java
felixdesouza: Fair enough

src/test/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
iamdanfox: Imagine you happened to have some containers running before your run this test. You'd get a failure in this test and dive into the code to start debugging.  You'd spend ages on this, until you eventually realise you just had a few containers up to start with and you'd just wasted hours of your life, so you add in one quick sanity check beforehand to save the next person ;)

src/test/java/com/palantir/docker/compose/connection/ContainerNamesShould.java
iamdanfox: Moved these all away

PR #117, opened 2016-10-14
New `circleAwareLogDirectory` enables partial log capture on CI

iamdanfox:
I've been using this little utility method to debug some tricky circle CI builds, so thought it's worth contributing back.

### DIFF
--- a/src/test/java/com/palantir/docker/compose/logging/LogDirectoryTest.java
+++ b/src/test/java/com/palantir/docker/compose/logging/LogDirectoryTest.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package com.palantir.docker.compose.logging;
+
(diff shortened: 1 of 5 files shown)


src/test/java/com/palantir/docker/compose/logging/LogDirectoryTest.java
```diff
@@ -0,0 +1,40 @@
...
+import static org.hamcrest.Matchers.is;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.EnvironmentVariables;
+
+public class LogDirectoryTest {
+
+    @Rule
+    public final EnvironmentVariables variablesRule = new EnvironmentVariables();
```
joelea: I didn't know this existed, nice.

iamdanfox: Yeah it's not an official JUnit thing, but it does the job :)

PR #121, opened 2016-10-18
Add string version of circleAwareLogDirectory.

SerialVelocity:
This solves the case when multiple classes have the same name, or the rule is part of a superclass


src/main/java/com/palantir/docker/compose/logging/LogDirectory.java
```diff
@@ -25,6 +25,11 @@ public static String circleAwareLogDirectory(Class<?> testClass) {
         return artifactRoot + "/dockerLogs/" + testClass.getSimpleName();
```
iamdanfox: Now that you've added the string one, what about passing using it here...

```java
return circleAwareLogDirectory(testClass.getSimpleName());
```

less code duplication!

PR #122, opened 2016-10-26
Fix errors that occur when containers are started/stopped multiple times

SerialVelocity:


iamdanfox:
Hmm, I think this might be a bit over-naive... the memoized ports was intentionally added for perf

iamdanfox:
@SerialVelocity have you found an alternative solution to the problems you mention above?  (Should we keep this PR open or can I close it out?)

SerialVelocity:
I ended up using reflection to override your code I think... So no, not a proper solution. @joelea, do you know whether the perf was super bad without the memoizing?

hpryce:
The perf using docker-machine was awful, maybe with Docker for Mac it's better but I don't believe anybody has tested it.

joelea:
I remember it being pretty bad - we just call ports so many freaking times. I can imagine us locally caching ports in places where they're used in a tight loop (e.g. when doing healthchecks) and still allow for the restarting.

Gui13:
I have the issue with the log collector here: a base test class that instanciates the DockerComposeRule, and it can only be used once because otherwise the log collector is not set to null correctly and cannot be restarted.
I don't have the problem of start/stop since my containers don't need to be stopped between tests.

I'd split this issue into two independent problems: 
- the log collector issue which is encountered even when there's no start/stop
- the port memoization issue, which is encountered when there's a start/stop in between tests.

joelea:
Closing due to failing tests + lack of activity. Happy to reopen the conversation if I'm missing some context.

PR #125, opened 2016-11-04
Switch to immutables strict builder

### LINKED ISSUE #118: .addAllClusterWaits overwrites previous waits.
```
ClusterWait a,b,c,d;
DockerComposeRule.builder()
  .addClusterWait(a)
  .addAllClusterWaits(ImmutableList.of(b,c,d)
```

Removes cluster wait 'a' from the list of waits. Things then fail in a very non transparent manner.

iamdanfox:
This removes a few generated methods too, e.g, you can no longer initialise a builder `from` an existing instance.

fixes #118

### DIFF
--- a/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
+++ b/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
@@ -43,7 +43,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @Value.Immutable
-@Depluralized
+@CustomImmutablesStyle
 public abstract class DockerComposeRule extends ExternalResource {
     public static final Duration DEFAULT_TIMEOUT = Duration.standardMinutes(2);
     public static final int DEFAULT_RETRY_ATTEMPTS = 2;
@@ -219,6 +219,10 @@ public abstract class DockerComposeRule extends ExternalResource {
         public Builder waitingForHostNetworkedPort(int port, HealthCheck<DockerPort> healthCheck, ReadableDuration timeout) {
             ClusterHealthCheck clusterHealthCheck = transformingHealthCheck(cluster -> new DockerPort(cluster.ip(), port, port), healthCheck);
             return addClusterWait(new ClusterWait(clusterHealthCheck, timeout));
+        }
+
+        public Builder clusterWaits(Iterable<? extends ClusterWait> elements) {
+            return addAllClusterWaits(elements);
         }
     }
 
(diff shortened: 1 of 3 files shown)


iamdanfox:
@jamesneale does this look OK to you?  This guarantees that there's no way to overwrite any previously set builder fields and adds a nicer convenience method to add more `clusterWaits`

src/test/java/com/palantir/docker/compose/DockerComposeRuleShould.java
```diff
@@ -265,6 +277,32 @@ public void when_remove_conflicting_containers_on_startup_is_set_to_false_before
...
+                mock(ClusterWait.class),
+                mock(ClusterWait.class));
+        Collection<ClusterWait> secondAssignment = ImmutableList.of(mock(ClusterWait.class));
+
+        DockerComposeRule twoAssignments = defaultBuilder()
+                .clusterWaits(originalAssignment)
+                .clusterWaits(secondAssignment)
+                .build();
+
+        assertThat(twoAssignments.clusterWaits(), hasSize(3));
```
jamesneale: Although in theory 3 is sufficient we should probably test what the test says we are xD

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -43,7 +43,7 @@
 import org.slf4j.LoggerFactory;
 
 @Value.Immutable
-@Depluralized
+@CustomImmutablesStyle
```
jamesneale: This is mega although name could probably be better

jamesneale:
LGTM some minor nits but happy to be disagreed with

src/test/java/com/palantir/docker/compose/DockerComposeRuleShould.java
```diff
@@ -224,9 +231,14 @@ public void be_able_to_save_logs_to_a_directory_while_containers_are_running()
 
     @Test
     public void not_shut_down_when_skipShutdown_is_true() throws InterruptedException {
-        defaultBuilder().skipShutdown(true)
-                        .build()
-                        .after();
+        DockerComposeRule.builder()
```
iamdanfox: `skipShutdown` only takes a boolean, but we are setting `shutdownStrategy` a bit unnecessarily.  Have scoped it down to the one test method that needs it.

src/main/java/com/palantir/docker/compose/DockerComposeRule.java
iamdanfox: Any suggestions?

Issue #44, opened 2016-04-05
saveToLogs does not save logs


verity:
V0.5.3

When using the following class rule, it creates the log directory, but never writes any container log files.

`@ClassRule
    public static final DockerComposition composition = DockerComposition.of("src/test/resources/full-stack/docker-compose.yml").saveLogsTo("build/docker-logs/dockerSuite").build()`

Debugging the code, I can see that in FileLogCollector.java > startCollecting
 `ContainerNames containerNames = dockerCompose.ps();` always returns empty.

The docker ps output contains a result, however `ContainerNames.parseFromDockerComposePs(psOutput)` is empty.

ContainerNames.java -> getContainerNamesAtStartOfLines seems to be the cause.

The psContainerOutput value it is trying to parse is:

> gaia       service/bin/gaia server va ...   Up      0.0.0.0:17000->17000/tcp, 0.0.0.0:17001->17001/tcp, 8000/tcp, 8001/tcp, 8443/tcp 
> selenium   /opt/bin/entry_point.sh          Up      0.0.0.0:4444->4444/tcp, 0.0.0.0:5900->5900/tcp

verity:
Upgrading the docker-compose file to V2 fixed this issue for me.

joelea:
Is this linked to #38 ?

CRogers:
@verity did originally use `container_name`s in her docker-compose file - just single words with no underscores iirc. This probably also triggered the bug, as there where no underscores for DCR to look for.

joelea:
yeah, this would've triggered it. The underscore bug was separate from this whole concept. I like that people tend to hit bugs in big clumps.

CRogers:
Basically, the container name in `docker-compose ps` can be anything.

joelea:
Yep, it's the same as #38, we'll need to pretty aggressively change how we grab container names.

joelea:
Fixed!

Issue #38, opened 2016-03-26
DockerCompose.ps() incorrectly ignores certain valid container names


j-baker:
To parse results, it calls https://github.com/palantir/docker-compose-rule/blob/505286255bee81cec0f537e1cb7a09aafb21e570/src/main/java/com/palantir/docker/compose/connection/ContainerNames.java#L48

which assumes that the container name is in the format PROJECT_name_SOMENUMBER, I think.

But, you can set the container name to whatever you want, and the name can also have underscores in.

In these cases, the container name will be filtered out.

For example, if the result of docker-compose ps is

```
47aab4ab_salt_config_1       /true                            Exit 0                                                              
cassandra.palantir.dev       /docker-entrypoint.sh cass ...   Up       7000/tcp, 7001/tcp, 7199/tcp, 9042/tcp, 9160/tcp           
```

then no containers will be returned.

joelea:
Loving the name change on the issue :+1: Though I do like the old name better.

This sucks and doesn't have an easy fix, I have a few ideas I'm going to let stew though. If you have any ideas, feel free to mention them or make a PR.

felixdesouza:
I was looking this up at some point. I think terminology needs to be cleared up also. 

You can list the the services (what you specify in your docker-compose file, regardless of `container_name` and also project name) that your compose file provides. If you're not doing any sort of scaling then there is a 1-1 mapping of docker-compose services (as defined by docker) and a container that runs an instance of that service. You can do this with:
```
docker-compose config --services
```
This requires docker-compose version 1.6 and I guess transitively docker 1.10 (although not entirely sure)

Not entirely sure what the scope is for supporting scaling especially within JUnit code, but suppose we didn't support scaling via `docker-compose scale service1=<num-s1> service2=<num-s2> ...` then we could do something like:

```
for service in 'docker-compose config --services':
    psOutput = docker-compose ps service
    parse container name(psOutput)
```

Regarding name clarification, I think our definition of a service may or may not be confusing given how docker-compose has its own concept.

As an aside, I was also thinking, as opposed to having to do a lot of parsing and regexing ourselves, once we know which containers are attached to which service, we could just connect to the docker API, which is just a HTTP endpoint. This would give us anything that we need to parse to infer state such as port mappings. It's not directly related, but fixing the issue in the way above allows this to be possible.

joelea:
oooh, docker-compose config --services is nice. That looks like the right kind of path to go down. Thanks1

[...]

PR #127, opened 2016-11-04
Use docker-compose config --services …

dev-kpyc:
instead of ps for getting the names of services to collect logs for.

https://github.com/palantir/docker-compose-rule/issues/38
https://github.com/palantir/docker-compose-rule/issues/44


joelea:
Well this seems like a delightfully sane way of dealing with those problems. Thanks!

PR #130, opened 2016-11-10
Revert incorrect change to LICENSE

nmiyake:
The brackets are in the section on re-applying the license, so it is correct to have them.


nmiyake:
Sorry about that (^^;)

PR #132, opened 2016-11-21
LocalBuilder: Allow additionalEnvironment to override systemEnvironment

### LINKED ISSUE #131: DockerMachine fails if "system" and "additional" environment variables are duplicated.
Relevant [code](https://github.com/palantir/docker-compose-rule/blob/master/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java#L100):
```
            Map<String, String> environment = ImmutableMap.<String, String>builder()
                    .putAll(systemEnvironment)
                    .putAll(additionalEnvironment)
                    .build();
```

If you set FOO=bar in the "additional environment", but FOO=bar happens to also be set on the system, this will fail.

Options:
1 - merge the entries, fail if there's a conflict (e.g. FOO=bar, FOO=baz -> fail)
2 - merge the entries, choose additional if there's a conflict (e.g. system FOO=bar, add. FOO=baz -> FOO=baz).

gsheasby:
The docker reserved variables (DOCKER_TLS_VERIFY, DOCKER_HOST,
DOCKER_CERT_PATH) are still not allowed in additionalEnvironment.
But the user can now override their own environment variables
programatically. (See AtlasDB PR 1237).

Fixes #131.

### DIFF
--- a/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
+++ b/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
@@ -22,7 +22,6 @@ import static com.palantir.docker.compose.configuration.EnvironmentVariables.DOC
 import static com.palantir.docker.compose.configuration.EnvironmentVariables.DOCKER_TLS_VERIFY;
 
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
 import com.palantir.docker.compose.configuration.AdditionalEnvironmentValidator;
 import com.palantir.docker.compose.configuration.DockerType;
 import com.palantir.docker.compose.configuration.RemoteHostIpResolver;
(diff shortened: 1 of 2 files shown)


src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
```diff
@@ -100,9 +101,11 @@ public LocalBuilder withEnvironment(Map<String, String> newEnvironment) {
         public DockerMachine build() {
             dockerType.validateEnvironmentVariables(systemEnvironment);
             AdditionalEnvironmentValidator.validate(additionalEnvironment);
+            Map<String, String> combinedEnvironment = Maps.newHashMap();
+            combinedEnvironment.putAll(systemEnvironment);
+            combinedEnvironment.putAll(additionalEnvironment);
             Map<String, String> environment = ImmutableMap.<String, String>builder()
-                    .putAll(systemEnvironment)
-                    .putAll(additionalEnvironment)
+                    .putAll(combinedEnvironment)
```
felixdesouza: The two maps seem redundant/verbose. I propose you either return the `combinedEnvironment` or do `ImmutableMap.copyOf(combinedEnvironment)`.

Issue #119, opened 2016-10-15
Prints something about circle ci when running locally


j-baker:
```7:53:16.426 [Thread-1] WARN  c.p.d.c.e.AggressiveShutdownStrategy - Couldn't shut down containers due to btrfs volume error, see https://circleci.com/docs/docker-btrfs-error/ for more info.```

this seems bogus since I'm not running in Circle CI.

**Comment by j-baker** [comment]
It did it again!

```
[2016-12-04 03:49:51]     Step 1 : FROM selenium/standalone-chrome-debug:3.0.1-aluminum
[2016-12-04 03:49:51]     Service 'chrome' failed to build: Get https://registry-1.docker.io/v2/selenium/standalone-chrome-debug/manifests/3.0.1-aluminum: dial tcp: lookup registry-1.docker.io on 172.17.87.1:53: no such host
[2016-12-04 03:49:51] 03:49:51.810 [Thread-5] INFO  c.p.d.c.e.AggressiveShutdownStrategy - Shutting down []
[2016-12-04 03:49:51] 03:49:51.885 [Thread-5] WARN  c.p.d.c.e.AggressiveShutdownStrategy - Couldn't shut down containers due to btrfs volume error, see https://circleci.com/docs/docker-btrfs-error/ for more info.```

j-baker:
(the dial tcp thing is a go bug, I think - but it's not the btrfs error)

PR #133, opened 2016-11-25
Cleanup networks after aggressive shutdown

### LINKED ISSUE #120: Aggressive shutdown mode doesn't seem to clean up networks that were created
It seems to exhaust available networks really quickly. Probably worth following up slaying all the containers with a call to docker-compose down.

ashrayjain:
This add a call to docker down after aggressive shutdowns.

Fixes #120

### DIFF
--- a/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
+++ b/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
@@ -27,25 +27,15 @@ public class AggressiveShutdownStrategy implements ShutdownStrategy {
         List<ContainerName> runningContainers = rule.dockerCompose().ps();
 
         log.info("Shutting down {}", runningContainers.stream().map(ContainerName::semanticName).collect(toList()));
-        if (removeContainersCatchingErrors(rule, runningContainers)) {
-            return;
-        }
-
-        log.debug("First shutdown attempted failed due to btrfs volume error... retrying");
-        if (removeContainersCatchingErrors(rule, runningContainers)) {
-            return;
-        }
-
-        log.warn("Couldn't shut down containers due to btrfs volume error, "
-                + "see https://circleci.com/docs/docker-btrfs-error/ for more info.");
+        removeContainersCatchingErrors(rule, runningContainers);
+        removeNetworks(rule);
     }
 
-    private boolean removeContainersCatchingErrors(DockerComposeRule rule, List<ContainerName> runningContainers) throws IOException, InterruptedException {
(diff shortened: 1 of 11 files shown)


iamdanfox:
:+1:

ashrayjain:
strange. that test passes locally for me. the network didn't get cleared up on circle somehow.

ashrayjain:
Ok, so this also fixes #119 (sort of) since it looks like the brtfs thing on circle throws an error while actually removing the containers (according to `docker-compose ps` at least). Given that, I think it might make sense to get rid of the retry stuff?

src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
```diff
@@ -27,25 +27,15 @@ public void shutdown(DockerComposeRule rule) throws IOException, InterruptedExce
...
-        }
-
-        log.debug("First shutdown attempted failed due to btrfs volume error... retrying");
-        if (removeContainersCatchingErrors(rule, runningContainers)) {
-            return;
-        }
-
-        log.warn("Couldn't shut down containers due to btrfs volume error, "
-                + "see https://circleci.com/docs/docker-btrfs-error/ for more info.");
+        removeContainersCatchingErrors(rule, runningContainers);
```
gsheasby: We now only attempt to remove containers once, where previously we had a second try.

iamdanfox: I don't think we can get rid of the retry logic in Aggressive.  The BTRFS retries were added in https://github.com/palantir/docker-compose-rule/pull/115 because Circle occasionally fails to shut down containers.  If we don't retry, then leftover containers will conflict with later tests.

Could we either: 
- Keep the retries in ShutdownStrategy.AGGRESSIVE
- Add the network behaviour to a new ShutdownStrategy (leaving the current aggressive unchanged)
- Or, use your custom ShutdownStrategy on a few circle builds (before merging into DCR) and convince us that the BTRFS retries aren't needed

ashrayjain: We could choose to keep the retries but I think we might not need them anymore so i would like to try running a few build with this change to verify that this is indeed the case (maybe circle fixed something on their end?).

iamdanfox: Cool.  Are you happy to copy your proposed ShutdownStrategy into your repo for a few builds?

ashrayjain:
Hit a flake (I think) on the first build but passed on retry. @iamdanfox does this look ok?

PR #143, opened 2017-01-27
Fix publishing variables

### LINKED ISSUE #141: 0.29.0 didn't publish
you forgot to set a version on the subprojects, so it was unspecified, so it didn't publish (but the build succeeded) :(

https://circleci.com/gh/palantir/docker-compose-rule/434

joelea:
@j-baker this should fix #141 right?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,9 +23,6 @@ repositories {
     jcenter()
 }
 
-group = 'com.palantir.docker.compose'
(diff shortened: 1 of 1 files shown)


iamdanfox:
I think this looks good, but a guaranteed way would be to add the `mavenLocal()` repository and then run `./gradlew publishToMavenLocal` and inspect the results.

PR #148, opened 2017-02-09
Support native Docker healthchecks

chrisalice:
Docker now natively supports healthchecks specified [in the docker image](https://docs.docker.com/engine/reference/builder/#/healthcheck) or [in the docker-compose config](https://docs.docker.com/compose/compose-file/#/healthcheck). docker-compose-rule will now wait for these healthchecks to pass before letting the test run. This allows users to fold their healthchecks into their Docker container specification, letting the team that releases the Docker image determine what makes their service healthy, rather than having each downstream product reinvent the wheel.


chrisalice:
Closing and reopening as #155.

PR #149, opened 2017-02-09
Add option to accept all certificates

diamondT:
Adds a builder method that installs a trust-all TrustManager and HostVerifier. 
Could be useful in tests that involve SSL communication with self-signed certificates. 

E.g.:
```
@ClassRule
public static DockerComposeRule docker = DockerComposeRule.builder()
            .file("src/test/resources/docker-compose.yml")
            .acceptAllCerts()
            .build();
```

### DIFF
--- a/docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
+++ b/docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
@@ -38,6 +38,12 @@ import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
 import java.util.List;
+import javax.net.ssl.HostnameVerifier;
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/docker-compose-rule](https://github.com/palantir/docker-compose-rule), @diamondT! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

hpryce:
This seems totally sensible to want to do in some tests, and in the main project I work on we do something very similar. However, we also do this outside of docker-compose-rule based tests and I don't think this is something that makes sense to invoke through docker-compose-rule especially as it happening as a side effect of a builder doesn't seem particularly clear to me.

This seems like something similar to https://github.com/palantir/docker-proxy-rule - useful when using docker-compose-rule but not intrinsically part of it. Is this something you want to try writing as your own tiny library? I'm sure others would use it if you published it on github.

PR #152, opened 2017-02-12
Look for "docker-compose" in path

diamondT:
Adds directories found in PATH to the list of possible paths for docker executables.
It would be nice to work out-of-the-box, i.e. without needing a special env variable (```DOCKER_LOCATION``` or ```DOCKER_COMPOSE_LOCATION```) if the executable(s) already exists in path, even if not in one of the well-known locations.

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
@@ -18,26 +18,40 @@ package com.palantir.docker.compose.execution;
 import static java.util.Arrays.asList;
 
 import java.io.File;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.StringUtils;
 
 public class DockerCommandLocations {
     private static final Predicate<String> IS_NOT_NULL = path -> path != null;
     private static final Predicate<String> FILE_EXISTS = path -> new File(path).exists();
 
     private final List<String> possiblePaths;
+    private final String exeName;
+    private final boolean lookInPath;
 
-    public DockerCommandLocations(String... possiblePaths) {
+    public DockerCommandLocations(String exeName, boolean lookInPath, String... possiblePaths) {
+        this.exeName = exeName;
+        this.lookInPath = lookInPath;
         this.possiblePaths = asList(possiblePaths);
     }
 
     public Optional<String> preferredLocation() {
+        List<String> envPath = (StringUtils.isNotBlank(exeName) && lookInPath)
+                ? Arrays.asList(System.getenv("PATH").split(File.pathSeparator)) :
+                Collections.emptyList();
 
-        return possiblePaths.stream()
-                .filter(IS_NOT_NULL)
-                .filter(FILE_EXISTS)
-                .findFirst();
+        return Stream.concat(
+                envPath.stream().map(path -> Paths.get(path, exeName).toAbsolutePath().toString()),
+                possiblePaths.stream())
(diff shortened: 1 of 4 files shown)


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
```diff
@@ -18,26 +18,40 @@
...
 public class DockerCommandLocations {
     private static final Predicate<String> IS_NOT_NULL = path -> path != null;
     private static final Predicate<String> FILE_EXISTS = path -> new File(path).exists();
 
     private final List<String> possiblePaths;
+    private final String exeName;
+    private final boolean lookInPath;
 
-    public DockerCommandLocations(String... possiblePaths) {
+    public DockerCommandLocations(String exeName, boolean lookInPath, String... possiblePaths) {
```
hpryce: I really like this idea, it should reduce the amount of configuration necessary for people to use the library. However, this constructor now seems to be getting quite complicated and I don't think it's possible to write a test for finding the executable in the path with how this is written.

This seems like a really good candidate for using Immutables (https://immutables.github.io/), which we use elsewhere in the codebase - see DockerComposeExecArgument for a simple example. Using immutables you could do something like this:

```
@Value.Immutable
public abstract class DockerCommandLocations {

  protected String executableName();
  protected List<String> additionalSearchLocations();

  @Value.Default
   protected List<String> path() {
      // read path environment variable
   }

  @Value.Derived
  public Optional<String> preferredLocation() { ... }

  public static Builder builder();

}
```

Then, using the builder, it is possible to write unit tests by setting the path @Value.Default value so we can ensure this continues working in the future.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
```diff
@@ -15,47 +15,53 @@
...
-    private final List<String> possiblePaths;
-    private final String exeName;
-    private final boolean lookInPath;
-
-    public DockerCommandLocations(String exeName, boolean lookInPath, String... possiblePaths) {
-        this.exeName = exeName;
-        this.lookInPath = lookInPath;
-        this.possiblePaths = asList(possiblePaths);
+    protected abstract String executableName();
+    protected abstract List<Optional<String>> additionalSearchLocations();
```
diamondT: ```additionalSearchLocations()``` has to be a list of optionals since ```System.getenv``` might or might not return a null value.
Alternatively, immutables optionally (sic) allows null values for plain lists but that would require an update to the latest version and wouldn't make a big difference imho.

chrisalice: Why not filter out the nulls?

diamondT: it's the immutables generated method that does not allow nulls

chrisalice: `Stream.of(...).filter(o -> o != null).collect(toList())` ought to do it

diamondT: agreed. however, having a signature of ```List<String>``` would fail in runtime (under certain circumstances). the ```Stream.of()``` solution is fine but the caller would have to know that they should call this method using this specific construct and not by using any other plain list.
putting the ```Optional``` in the signature ensures null-safety at compile time, right?

hpryce:
@diamondT sorry for the delay here. I've pulled these changes into https://github.com/palantir/docker-compose-rule/pull/175 to include support for Windows, all going well we should be able to ship them soon.

I'm going to close this PR out as the linked one includes everything this does.

diamondT:
@hpryce Cool, thanks for the update.

PR #155, opened 2017-02-14
Support native docker healthchecks

chrisalice:
Docker now natively supports healthchecks specified [in the docker image](https://docs.docker.com/engine/reference/builder/#/healthcheck) or [in the docker-compose config](https://docs.docker.com/compose/compose-file/#/healthcheck). docker-compose-rule will now wait for these healthchecks to pass before letting the test run. This allows users to fold their healthchecks into their Docker container specification, letting the team that releases the Docker image determine what makes their service healthy, rather than having each downstream product reinvent the wheel.

Unfortunately we cannot currently run tests of this new feature on CI due to #156.


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Container.java
```diff
@@ -90,28 +93,36 @@ public DockerPort port(int internalPort) {
...
     }
 
     public void kill() throws IOException, InterruptedException {
-        dockerComposeProcess.kill(this);
+        dockerCompose.kill(this);
     }
 
     public State state() throws IOException, InterruptedException {
-        return dockerComposeProcess.state(containerName);
+        String id = dockerCompose.id(this).orElse(null);
```
iamdanfox: Out of curiosity, why didn't you go for:

```java
dockerCompose.id(this).map(docker::state).orElse(State.DOWN)
```

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/ContainerCache.java
```diff
@@ -15,22 +15,33 @@
...
 
-    public ContainerCache(DockerCompose dockerCompose) {
+    public ContainerCache(Docker docker, DockerCompose dockerCompose) {
+        this.docker = docker;
         this.dockerCompose = dockerCompose;
     }
 
     public Container container(String containerName) {
-        containers.putIfAbsent(containerName, dockerCompose.container(containerName));
+        containers.putIfAbsent(containerName, new Container(containerName, docker, dockerCompose));
```
iamdanfox: All this wiring in of `docker`, `docker-compose` dependencies feels a bit weird - not sure I have a better idea though

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/State.java
```diff
@@ -15,22 +15,21 @@
...
 package com.palantir.docker.compose.connection;
 
-import com.google.common.base.Preconditions;
-import com.google.common.base.Strings;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 public enum State {
-    Up, Exit;
+    DOWN, PAUSED, UNHEALTHY, HEALTHY;
```
iamdanfox: https://docs.docker.com/engine/reference/builder/#/healthcheck says "This status is initially starting"... does this map to `UNHEALTHY` initially?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterHealthCheck.java
```diff
@@ -37,5 +43,26 @@ static ClusterHealthCheck serviceHealthCheck(String containerName, HealthCheck<C
         };
     }
 
-    SuccessOrFailure isClusterHealthy(Cluster cluster);
+    static ClusterHealthCheck nativeHealthChecks() {
```
iamdanfox: Would it be feasible to put some logging in here?  It's pretty frustrating to debug an existing healthcheck, so it would be cool if the new ones are better ;)

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterHealthCheck.java
```diff
@@ -37,5 +43,26 @@ static ClusterHealthCheck serviceHealthCheck(String containerName, HealthCheck<C
...
+        return cluster -> {
+            Set<String> unhealthyContainers = new LinkedHashSet<>();
```
iamdanfox: Personally I'd kinda prefer to Java8 this and do a stream, map thing - it emphasises there's no mutation or anything fancy happening here.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterHealthCheck.java
```diff
@@ -37,5 +43,26 @@ static ClusterHealthCheck serviceHealthCheck(String containerName, HealthCheck<C
...
+            try {
+                for (Container container : cluster.allContainers()) {
+                    State state = container.state();
+                    if (state == State.UNHEALTHY) {
```
iamdanfox: Perhaps worth capturing in a javadoc comment that this will not wait for any DOWN or PAUSED containers.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -165,14 +165,18 @@ private Version version() throws IOException, InterruptedException {
     }
 
     @Override
-    public List<String> services() throws IOException, InterruptedException {
-        String servicesOutput = command.execute(Command.throwingOnError(), "config", "--services");
-        return Arrays.asList(servicesOutput.split("\n"));
+    public Optional<String> id(Container container) throws IOException, InterruptedException {
+        String id = command.execute(Command.throwingOnError(), "ps", "-q", container.getContainerName());
```
iamdanfox: I wonder if we're gonna get perf hits from repeatedly calling this :/

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/connection/ContainerIntegrationTests.java
```diff
@@ -0,0 +1,79 @@
...
+        Container container = new Container("noHealthcheck", docker, dockerCompose);
+        assertEquals(State.DOWN, container.state());
+        container.up();
+        assertEquals(State.HEALTHY, container.state());
+        container.kill();
+        assertEquals(State.DOWN, container.state());
+    }
+
+    /**
+     * This test is not currently enabled in Circle as it does not provide a sufficiently recent version of docker-compose.
```
iamdanfox: Is this javadoc up to date?  I don't see an @Ignore here.

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/connection/ContainerIntegrationTests.java
```diff
@@ -0,0 +1,79 @@
...
+    public void testStateChanges_withHealthCheck() throws IOException, InterruptedException {
+        assumeThat("docker version", Docker.version(), new GreaterOrEqual<>(Version.forIntegers(1, 12, 0)));
+        assumeThat("docker-compose version", DockerCompose.version(), new GreaterOrEqual<>(Version.forIntegers(1, 10, 0)));
+
+        DockerCompose dockerCompose = new DefaultDockerCompose(
+                DockerComposeFiles.from("src/test/resources/native-healthcheck.yaml"),
+                dockerMachine,
+                ProjectName.random());
+
+        // The withHealthcheck service's healthcheck checks every 100ms whether the file "healthy" exists
```
iamdanfox: v. cool

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/connection/ContainerIntegrationTests.java
iamdanfox: Aaah it's the assumeThat that disables this... got it.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/State.java
chrisalice: Yes.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Container.java
chrisalice: Because state throws checked exceptions

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterHealthCheck.java
chrisalice: Ditto: checked exceptions suck

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
chrisalice: It's pretty fast

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterHealthCheck.java
chrisalice: Done, thanks.

iamdanfox:
LGTM let's 🚢 it

PR #157, opened 2017-02-14
Speed up test teardown

chrisalice:
Most Docker containers ignore the SIGTERM they are sent by 'docker-compose down', resulting in a pointless 10 second delay per test. The 'aggressive' shutdown strategies are unnecessarily complex, however, as we can simply issue a 'docker-compose kill' first, and rely on 'docker-compose down' to clean up gracefully. There is also no need to issue a 'docker-compose rm' as we already pass '--volumes' to 'down'. This new KillDownShutdownStrategy should be safe for us to default users to.


iamdanfox:
:+1: that aggressive thing was a hail mary

PR #166, opened 2017-03-01
Fix docker health check to work on Windows.

samwright:
Using version 0.31.0 and the latest stable docker release (version 1.13.1) on Windows, both of the tests in ContainerIntegrationTests fail with the following error:
```
com.palantir.docker.compose.execution.DockerExecutionException: 'docker inspect --format={{if not .State.Running}}DOWN{{else if .State.Paused}}PAUSED{{else if index .State "Health"}}{{if eq .State.Health.Status "healthy"}}HEALTHY{{else}}UNHEALTHY{{end}}{{else}}HEALTHY{{end}} 0f0d3e40ef642d0b1201f23e50994d531c33e44fd7f05104dd111d98bdf0dd57' returned exit code 64
The output was:
Template parsing error: template: :1: function "Health" not defined
```

This is down to the way the double-quotes need to be escaped in windows, which is to use the back-tick. My commit performs this escaping when running on Windows.

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/Docker.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/Docker.java
@@ -25,6 +25,8 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.SystemUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


iamdanfox:
Awesome fix @samwright, this would have taken me a while to get right devving on a mac!

As an aside, it's a shame we don't have any windows tests that can guarantee this keeps working.  I've requested AppVeyor access for this repo and will try to get some kind of minimal windows CI going.  Do you want this released fast or would you be able to wait until we can put a test on this?

chrisalice:
No point waiting for a test — it can't make things worse, after all.

PR #169, opened 2017-03-01
Fix call to docker-compose exec in integration test.

samwright:
Calls to docker-compose exec in Windows need the -T option (to disable
pseudo-tty allocation, which doesn't work in Windows without something
like winpty).

This commit adds the -T option to all calls to docker-compose exec.

Is there any reason not to disable pseudo-tty allocation? If so I could only add "-T" if on Windows.

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
@@ -140,7 +140,11 @@ public class DefaultDockerCompose implements DockerCompose {
 
     private static String[] constructFullDockerComposeExecArguments(DockerComposeExecOption dockerComposeExecOption,
             String containerName, DockerComposeExecArgument dockerComposeExecArgument) {
+        // The "-T" option here disables pseudo-TTY allocation, which is not useful here since we are not using
+        // terminal features here (e.g. we are not sending ^C to kill the executed process).
+        // Disabling pseudo-TTY allocation means this will work on OS's that don't support TTY (i.e. Windows)
         ImmutableList<String> fullArgs = new ImmutableList.Builder<String>().add("exec")
+                                                                            .add("-T")
                                                                             .addAll(dockerComposeExecOption.options())
                                                                             .add(containerName)
                                                                             .addAll(dockerComposeExecArgument.arguments())
--- a/docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
+++ b/docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
@@ -172,7 +172,7 @@ public class DockerComposeShould {
             throws IOException, InterruptedException {
         when(executedProcess.getInputStream()).thenReturn(toInputStream("docker-compose version 1.7.0rc1, build 1ad8866"));
         compose.exec(options("-d"), "container_1", arguments("ls"));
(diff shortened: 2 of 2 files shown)


samwright:
Strange, CirclCI says the build was successful but it hasn't updated the status above...

chrisalice:
Reran; it worked this time

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -141,6 +141,7 @@ private Version version() throws IOException, InterruptedException {
     private static String[] constructFullDockerComposeExecArguments(DockerComposeExecOption dockerComposeExecOption,
             String containerName, DockerComposeExecArgument dockerComposeExecArgument) {
         ImmutableList<String> fullArgs = new ImmutableList.Builder<String>().add("exec")
+                                                                            .add("-T")
```
iamdanfox: Could we put a tiny comment in here to capture how this is necessary for Windows?  I can imagine someone stumbling across this in future and wondering what a magic capital T does and why their unit tests seem to pass without it!!

https://docs.docker.com/compose/reference/exec/

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
```diff
@@ -193,15 +193,15 @@ public void pass_concatenated_arguments_to_executor_on_docker_compose_run()
 
     @Test
     public void return_the_output_from_the_executed_process_on_docker_compose_exec() throws Exception {
-        String lsString = "-rw-r--r--  1 user  1318458867  11326 Mar  9 17:47 LICENSE\n"
-                + "-rw-r--r--  1 user  1318458867  12570 May 12 14:51 README.md";
+        String lsString = String.format("-rw-r--r--  1 user  1318458867  11326 Mar  9 17:47 LICENSE%n"
```
iamdanfox: Great `%n`! I bet there are a ton of places we need to fix this

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
samwright: Will do

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
samwright: There are indeed a lot of "\n" in test strings, but I think this is the only place that matters. Command::processOutputFrom(Process) reads the lines from `process.getInputStream()` and joins them using `System.lineSeparator()`, so lsString's "\n" gets read as "\r\n" on Windows, hence the problem here. The other places where "\n" is used (e.g. ContainerNameShould) just test that the linux line endings work. I added `ContainerNameShould.allow_windows_newline_characters` to test the windows line endings.

samwright:
@iamdanfox it seems I have the (dubious :-P) honour of being the first to go through AppVeyor validation. Would you mind taking a look? I think it needs reconfiguring.

iamdanfox:
Haha yep, so I've set AppVeyor as a non-required check, so we can still merge it.

screen shot 2017-03-02 at 15 30 32

To make the red cross go away, we'll just need to add an appveyor.yml file that is essentially a no-op (e.g. just `./gradlew dependencies`).  We can flesh out the full test suite later!

PR #170, opened 2017-03-01
Feature/use path to find executables

samwright:
This searches through the path variable's directories for the docker
commands, instead of looking in predetermined OS-specific locations or
requiring new environmental variables to be set.

This also renames DockerCommandLocations to DockerCommandLocator, which better
reflects what it is

It also fixes the findbugs excludeFilter.xml file so the bugs to ignore for test classes are actually ignored for the test classes.


hpryce:
@samwright I've pulled these changes into https://github.com/palantir/docker-compose-rule/pull/175 to include the other PR that was open for similar functionality.

Sadly it's more complicated that what you've got here, but Docker for Mac installs into a location that isn't on the path when running the tests through Eclipse on a Mac and the environment variable overrides are needed because we install an alternative (more recent) copy of docker-compose as part of CI and need a way to specify that.

I'm going to close this PR out in favour of the linked one.

PR #175, opened 2017-03-07
Feature/look in path

hpryce:
This combines https://github.com/palantir/docker-compose-rule/pull/170 and https://github.com/palantir/docker-compose-rule/pull/152 to provide support for finding docker from the path both on Windows and in existing setups.


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
```diff
@@ -1,47 +1,75 @@
 /*
  * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
felixdesouza: Are we keeping or removing the License? It's removed here but kept in the following file.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
```diff
@@ -1,47 +1,75 @@
...
+
+    private Stream<String> pathLocations() {
+        Stream<String> pathLocations = Stream.concat(PATH_SPLITTER.splitAsStream(path()), macSearchLocations());
+        if (locationOverride() == null) {
+            return pathLocations;
+        }
+        return Stream.concat(Stream.of(locationOverride()), pathLocations);
+    }
+
+    public Stream<Path> forCommand() {
```
felixdesouza: Why are we returning `Stream`? They have use once semantics

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocator.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.commons.lang3.SystemUtils;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public abstract class DockerCommandLocator {
+
+    protected abstract String command();
+
+    @Value.Default
+    protected boolean isWindows() {
```
felixdesouza: returning a constant to represent a default where the constant is framed as a question is somewhat misleading (I still don't know what the default is)

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerComposeExecutable.java
```diff
@@ -29,19 +29,12 @@
...
-            "/usr/local/bin/docker-compose",
-            "/usr/bin/docker-compose"
-    );
-
     private static String defaultDockerComposePath() {
-        String pathToUse = DOCKER_COMPOSE_LOCATIONS.preferredLocation()
-                .orElseThrow(() -> new IllegalStateException(
-                        "Could not find docker-compose, looked in: " + DOCKER_COMPOSE_LOCATIONS));
-
+        DockerCommandLocator commandLocator = DockerCommandLocator.forCommand("docker-compose")
```
felixdesouza: returning a builder just to do this looks kinda meh, might as well do the entire builder from here and delete the `forCommand` method

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
hpryce: All files should have the Apache licence. I'll fix the ones that had it removed.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocator.java
hpryce: The default is to allow it to be overriden in tests, otherwise it should behave as a constant.

Would "useWindowsExecutableNaming" have been easier to understand?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
hpryce: Purely because it was extracted out the middle of a stream, I'll make it a list.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocator.java
felixdesouza: `isWindows` is fine, but I'm starting to think that making it a default doesn't really make any sense. Whoever creates this builder will know at the time what environment they're in, and it feels odd to be like "I'm not going to specify if I'm running on windows because defaults"

Unless of course this is for backcompat and it's used in client code then nevermind.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
```diff
@@ -1,47 +1,75 @@
...
-    public DockerCommandLocations(String... possiblePaths) {
-        this.possiblePaths = asList(possiblePaths);
+    @Value.Default
+    protected Map<String, String> env() {
+        return System.getenv();
     }
 
-    public Optional<String> preferredLocation() {
+    @Value.Derived
+    protected String path() {
```
samwright: This doesn't work on Windows, where the path variable is "Path". You could add another `if (path == null) path = env().get("Path")`, or iterate through the env() map to find any key that equals "path" ignoring case, e.g.

```
        String path = getEnv().entrySet().stream()
                .filter(e -> e.getKey().equalsIgnoreCase("path"))
                .findFirst()
                .map(Map.Entry::getValue)
                .orElseThrow(() -> new IllegalStateException("Could not find path variable in env"));
```

hpryce: Okay, this should now cover PATH, path and Path.

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerCommandLocatorShould.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test public void
+    provide_the_first_command_location_on_windows() {
+        DockerCommandLocator locator = DockerCommandLocator.forCommand(command)
+                .isWindows(true)
+                .build();
+        assertThat(locator.getLocation(), is(windowsCommandFileLocation));
+    }
+
+    @Test public void
+    fail_when_no_paths_contain_command() {
```
samwright: This is no longer true

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocations.java
```diff
@@ -13,35 +13,36 @@
...
-                .filter(IS_NOT_NULL)
-                .filter(FILE_EXISTS)
-                .findFirst();
     }
 
-    @Override
-    public String toString() {
-        return "DockerCommandLocations{possiblePaths=" + possiblePaths + "}";
+    public static List<Path> pathLocations() {
+        String path = Stream.of(System.getenv("PATH"), System.getenv("path"), System.getenv("Path"))
```
samwright: But what about CrazyOS that uses "pATh"? /s

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCommandLocator.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Value.Derived
+    protected List<Path> searchLocations() {
+        if (locationOverride() == null) {
+            return DockerCommandLocations.pathLocations();
+        }
+        return singletonList(Paths.get(locationOverride()));
+    }
+
+    public String getLocation() {
```
samwright: Previously DOCKER_COMPOSE_LOCATION would be the path to the binary, e.g. `/path/to/docker-compose`, but the logic here requires it to point to the directory in which to find docker-compose, i.e. `/path/to`. I think the previous behaviour is more obvious/expected.

You could remove the `searchLocations` method, and do the check for the location override here instead, e..g

```
if (locationOverride() == null) {
  return DockerCommandLocations.pathLocations().stream()....;
} else {
  return locationOverride();
}
```

It might even be worth moving `DockerCommandLocations.pathLocations()` to this class, but that's just a design thing.

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerCommandLocatorShould.java
samwright: Sorry, ignore this one. I got an error running this but I now can't recreate it.

samwright: Ah, found it again (I had @Ignore'd the test, woops!)

`Expected: (an instance of java.lang.IllegalStateException and exception with message a string containing "Could not find not-a-real-command! in [/usr/local/bin, /usr/bin]")
     but: exception with message a string containing "Could not find not-a-real-command! in [/usr/local/bin, /usr/bin]" message was "Could not find not-a-real-command! in [\usr\local\bin, \usr\bin]"`

The "/usr/local/bin" string is converted to a `Path` in `DockerCommandLocations.pathLocations()`, which on Windows converts it to "\usr\local\bin" (aren't path separators fun?!)

You could change the expected message to just `"Could not find " + command`, or create a `List<Path> defaultPaths = ImmutableList.of(Paths.get("/usr/local/bin"), Paths.get("/usr/bin"))` and expect `"Could not find " + command + " in " + defaultPaths`.

hpryce:
Okay, it turns out ProcessBuilder looks for commands on the path (at least on Mac/Linux) and so there is no need to do this manually. I've updated the PR to do the simpler thing of only looking in the known Mac install locations and allow for explicitly setting it through an environment variable.

@samwright could you give this a spin and see if it is actually sufficient for Windows? If not we should find a library that will give us as because as @chrisalice pointed out the Windows path variable can contain expansions that would require us to look at other environment variables to correctly use.

samwright:
Yep this works in Windows. I also tried just running "docker-compose.exe" without specifying its location and that worked too:

```
ProcessBuilder pb = new ProcessBuilder("docker-compose.exe", "-v");
Process p = pb.start();
p.waitFor();
BufferedReader r = new BufferedReader(new InputStreamReader(p.getInputStream(), StandardCharsets.UTF_8));
System.out.println(r.lines().collect(Collectors.toList())); // Prints: docker-compose version 1.11.2, build f963d76f
```

So it would probably be best to get rid of all that path variable stuff. Sorry for giving you a red herring!

dotCipher:
@hpryce , is this still relevant / needed to merge? Or is this overcome-by-events?

nedtwigg:
On windows, the just-released `0.34.0` is giving me `Could not find docker-compose`, even though `docker-compose.exe` is right on the path.  Seems like this PR would fix it.

PR #177, opened 2017-03-08
Add support for new YY.MM docker version format.

samwright:
This changes the regex for parsing the docker version to support the new YY.MM format (e.g. 17.03.0-ce). It also adds a non-static method in Docker for checking its version to allow for easier testing.


samwright:
Could someone try triggering another build on CircleCI? I have a suspicion that the failing test `FileLogCollectorShould.collect_logs_when_one_container_is_running_and_does_not_terminate_until_after_start_collecting_is_run` has a race condition.

iamdanfox:
@samwright have triggered.  Also if you want to fire off another one yourself, I think you can just amend your commit (with no changes) and then force push ;)

PR #180, opened 2017-03-09
Feature/windows smoke tests

hpryce:


gradle/integTest.gradle
```diff
@@ -0,0 +1,67 @@
+// From https://raw.githubusercontent.com/spring-io/sagan/master/gradle/integTest.gradle
```
iamdanfox: Please use https://github.com/unbroken-dome/gradle-testsets-plugin instead of copying and pasting gradle!!

hpryce:
Right now this is blocked on getting the latest version of docker-compose - https://chocolatey.org/packages/docker-compose doesn't have the right version yet.

Issue #202, opened 2017-06-16
Can't connect to DB after isListeningNow succeed


kasecato:
* version: 0.32.0
* db: MySQL 8
* junit: 4.12
* docker: 17.03.1
* os: macOS Sierra
* reproduction repository: https://github.com/k--kato/docker-compose-rule-db-parallel

I cannot connect to MySQL even after `isListeningNow` succeeded.  How can I wait until JDBC can connect?

```java
    private DockerComposeRule dockerDb;
    private Connection conn;

    @Before
    public void before() throws IOException, InterruptedException, SQLException {
        dockerDb = DockerComposeRule.builder()
                .file("src/test/resources/docker-compose-test.yml")
                .waitingForService("db", HealthChecks.toHaveAllPortsOpen())
                .addClusterWait(new ClusterWait(
                        ClusterHealthCheck.transformingHealthCheck(
                                cluster -> cluster.container("db").port(3306),
                                target -> SuccessOrFailure.fromBoolean(target.isListeningNow(), target + " was not opened")),
                        Duration.standardSeconds(30)))
                .build();

        dockerDb.before();

        final MysqlDataSource dataSource = new MysqlDataSource() {{
            final String url = dockerDb.containers().container("db").port(3306).inFormat("jdbc:mysql://$HOST:$EXTERNAL_PORT/sampledb");
            setURL(url);
            setUser("root");
            setPassword("sa");
        }};

        conn = dataSource.getConnection();
        // I don't want to use like this codes
/*
        while (true) {
            try {
                conn = dataSource.getConnection();
                break;
            } catch (Exception e) {
                Thread.sleep(1000);
            }
        }
*/
    }

    @After
    public void tearDown() {
        dockerDb.after();
    }

    @Test
    public void test1() throws SQLException {
        conn.prepareStatement("SELECT 1").executeQuery();
}
```

dotCipher:
I already wrote a `JdbcContainerHealthCheck` for my project, was originally going to submit a PR back here, but was unsure if it would be appropriate for tracking individual Health check types?

Would it be fair to have it under `HealthChecks` ? Or is it too specific of a use case?

Either way I would be happy to contribute it back.

k--kato:
@dotCipher  I'd love to use `JdbcContainerHealthCheck`. Is it possible to paste code here?

dotCipher:
@iamdanfox / @hpryce for input here, but I think the most scalable solution would be to commit back to the product instead of pasting code around.

[...]

PR #208, opened 2017-08-16
Extra Health Checks (with first being JDBCContainerHealthCheck)

dotCipher:
## Highlights
- Added gradle submodule for extra health-checks
- Added `JdbcContainerHealthCheck` for handling appropriate database readiness signal
   - Includes convenience constructors for `postges` and `mysql`, but more could be added fairly trivially.
   - Will fail fast on `JdbcContainerConfiguration` construction, based on using the `DriverManager` API.  It will interpret if the configured JDBC Url can be handled by any of the current `Driver`s loaded in the runtime, before a docker `Container` is ever started.  This saves time from having to wait for the default to `ClusterWait` to timeout.

## RFC
@iamdanfox


dotCipher:
@iamdanfox I might be missing something but is circle being finicky here?

yiweig:
+1 on getting this feature in

euclio:
+1, just ran into a flaky test due to this issue.

dotCipher:
Not sure if we should encourage users to use the solution posted by `mdotson` on #202 
(example:
```
healthcheck:
  test: ["CMD", "psql", "-h", "localhost", "-c", "SELECT version();"]
  interval: 1s
  timeout: 5s
  retries: 12
```
)

Or we go with this codified option.  @iamdanfox, any thoughts?

kasecato:
@dotCipher @dotson I don't think docker `healthcheck` and `jdbc` are the same. I tried it both, but your PR jar succeeded in testing and docker healthcheck fails in my test.

dotCipher:
Interesting note @kasecato , that might mean that we should try and get this merged.  Apologies on the original radio silence

dotCipher:
@iamdanfox would you be able to take a look at the errors on circle here?  It seems like the version of docker engine that circle is running isn't what compose expects?

Would we be able to get this merged?

chrisalice:
As mentioned on slack, the build error is because public CircleCI doesn't support docker-compose files at version 3.1.

chrisalice:
I strongly recommend trying to get a more representative docker healthcheck coded, rather than merging this, as then you'll not only get docker-compose-rule working, but any other healthcheck-using systems can use the result (e.g. docker ps!)

dotCipher:
@alicederyn I am still a bit confused as to what you mean by "more representative docker healthcheck"?  Can you elaborate a bit more?

alicederyn:
Kasecato wrote "I don't think docker healthcheck and jdbc are the same" -- but the Docker healthcheck is an arbitrary script so it can be anything we want. We just need to write one that better matches jdbc.

dotCipher:
Closing due to staleness

PR #212, opened 2017-10-04
Better error logging

CRogers:
This way we print out what's actually going wrong with the connection rather than "it just didn't work"

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/Exceptions.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/Exceptions.java
@@ -1,0 +1,14 @@
+package com.palantir.docker.compose.connection.waiting;
+
+import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/Exceptions.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.docker.compose.connection.waiting;
+
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+
+public enum Exceptions {
+    ;
```
joelea: :(

CRogers: Hey, the "organically reared zero variant enum with side of flambé hanging semicolon" is a classic ~~dish~~coding style!

PR #223, opened 2018-01-27
JUnit5 Support

dotCipher:
_(NOTE: This comment is out-of-date and does not reflect the current state of this feature, see the most recent **Specification Overview** comment further down this thread for accurate information)_

### Overview
~~Provide basic functionality that replicates junit4's `@ClassRule` lifecycle by implementing an [`Extension`](http://junit.org/junit5/docs/current/user-guide/#extensions) for [`BeforeAllCallback`](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/BeforeAllCallback.html) and [`AfterAllCallback`](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/AfterAllCallback.html) dubbed `DockerComposeExtension` that uses reflection on the junit5 class to manipulate docker compose / machine state.  Basic expected usage is outlined below.~~

### Usage
```java

import org.junit.jupiter.api.extension.ExtendWith;
import com.palantir.docker.compose.DockerComposeExtension;

@ExtendWith(DockerComposeExtension.class)
class Junit5Tests {

     // The DockerComposeExtension will reflect on static fields for Junit5Tests
     // and control state as class container starts (before all tests) and when class container stops (after all tests)
     static final DockerComposeRule DOCKER_COMPOSE = DockerComposeRule.builder()
                                   // All the usual stuff
                                   .build();

     @Test
     void testThing() {
          // Standard test
     }

     @ParameterizedTest
     void testParameterized(String input) {
           // Parameterized test bound to the test class container lifecycle
     }

     @TestFactory
     Stream<DynamicTest> testFactory() {
          // Also should support @TestFactory since it's also bound to the parent class container lifecycle
     }

}
```

### TODO
~~- [ ] Tests for user fail cases of `DockerComposeExtension`~~
~~- [ ] Tests for expected correct behaviour of `DockerComposeExtension` used with `DockerComposeRule`~~
~~- [ ] (maybe?) Implement a junit4 `@Rule` equivalent `Extension` that uses [`BeforeEachCallback`](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/BeforeEachCallback.html) and [`AfterEachCallback`]~~(http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/AfterEachCallback.html)
~~- [ ] Fix Builder pattern on `DockerComposeRule` (Would love to get feedback on this, since I keep getting this error when trying to replicate the current Builder in junit4 version:~~

screen shot 2018-01-26 at 5 49 07 pm


dotCipher:
@chrisalice / @iamdanfox, in it's current state, this doesn't actually work.  But I wanted to get some eyes on it before I go any further in case anyone had opinions and/or thoughts on it's current direction.

Thanks!

dotCipher:
Tracking changes towards fixing #224 here

chrisalice:
How do I call methods on the docker containers?

chrisalice:
Can we use meta-annotations to drive DCR with a single DCR-specific annotation in the common case? e.g.

```
    @DockerCompose(project="upgrades", containers = "foo.yml", logTo = "container_logs")
```

chrisalice:
> static final DockerComposeRule

That ain't no rule...

dotCipher:
Hey @chrisalice,

Thanks for the feedback, I realize that it isn't an actual Rule.. bad naming I guess, sorry about that, just wanted to keep the standard based on the repo name. 

To answer your first question with the current implementation, in order to call methods on the containers you could use the instance you created statically (ie `DOCKER_COMPOSE`) in the example above.

We don't need to have it annotation driven, but have it just be an instance or static field that the user initializes (See [registering extensions](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-static-fields)).  For example:
```java
class DockerStaticInstanceExample {

    @RegisterExtension
    static DockerComposeExtension dce = DockerComposeExtension.builder()
        .files(someFiles())
         //  etc
        .build();

    @Test
    void testThings() {
        // can call dce.containers() for access to containers, etc
        assertThat(dce.containers().ip()).isNotEmpty();
    }
}
```
This should handle both cases where a static declaration would instantiate it once for the entire test class, or an instance field would instantiate it once per test.

However if we want to shift the paradigm to being annotation driven instead, we could do that instead, but what kind of model would you propose a user gets access to the underlying `DockerCompose` object and it's containers etc? Injection or some other means?

chrisalice:
Have another property, say, `teardownBetweenTests=true`.

dotCipher:
Ok, that's doable.  So you are in sync with the injection workflow for accessing the containers of the DockerCompose, etc?

chrisalice:
Yes, that's a good solution.

dotCipher:
@chrisalice I am working up a solution for it now, I ran into a config point that would make sense being part of the annotation, but requires it to be an enum.  See #230 for info

dotCipher:
Still a WIP, do not merge.

dotCipher:
Closing due to staleness

PR #226, opened 2018-02-06
com.jayway.availability -> org.availability

jhoch-palantir:


qinfchen:
close in favor of https://github.com/palantir/docker-compose-rule/pull/423

PR #230, opened 2018-03-20
Convert ShutdownStrategy to enums

dotCipher:
Switched to using an enum for ShutdownStrategy in order to allow use within annotations. This should allow the `ShutdownStrategy` to be provided as an enum config point to the annotation for the Junit5 implementation of DCR in #223


dotCipher:
@chrisalice

dotCipher:
This should allow the `ShutdownStrategy` to be provided as an enum config point to the annotation for the Junit5 implementation of DCR in #223

dotCipher:
Chatted off-thread with @chrisalice and we decided to just hard-code the shutdown strategy for junit5's implementation to the default of KILL_DOWN.

Issue #200, opened 2017-05-31
Memoized port mappings cause issues when a service restarts


FrelliBB:
I am using docker-compose-rule in a test but encountered an issue that is giving me some trouble. I have a spring boot service which depends on a mysql database container to be ready first in order to successfully start up. To solve this issue I added a restart policy to my service so my compose file looks something like this (irrelevant lines omitted).

```yml
services:
  database:
    image: myregistry.com/database:latest

  my-service:
    links:
    - database:database.dev
    restart: unless-stopped
```

When the service restarts due to the restart policy the port mapping is changed. This means that the memoizing of portmappings in `com.palantir.docker.compose.connection.Container` will have the wrong port after the service restarts.

Would it be possible to remove this memoizing? Or is there a specific reason for it?

iamdanfox:
The memoizing stuff was added because the 'wait for services' stage was found to be very slow.  I think it would be good to remove the memoizing stuff and re-add it somewhere in our ClusterWait code.

Could you pop a PR up?  Hopefully should be a lot of deleted code :)

felixdesouza:
Alternatively, if you want to control startup order you can try some of the things listed here:
https://docs.docker.com/compose/startup-order/

It would allow your test classes to remain the same and not need any logic to restart (even though it's been up once).

I still agree the mappings are borked if a service restarts, but I would expect a service to restart in an exceptional/intentional case e.g. you want to see how your service handles failure etc. as opposed to startup dependencies.

fryz:
Think this is the same issue as https://github.com/palantir/docker-compose-rule/issues/163

vmykhailiuk:
@fryz Could somebody merge fix for this issue and release it? Please help.

fryz:
Hey @vmykhailiuk  - sorry for the delay. 

We've merged the PR and are working on cutting a release. 

Just a heads up, we are actually in the process of validating TestContainers as a replacement for docker-compose-rule. (see https://github.com/palantir/docker-compose-rule/issues/181 for context). 

So while this was pretty straightforward to merge and cut a release (thanks for the contribution @mktange!), we aren't going to be putting much, if any, effort into future development until we've validated the migration and, if things look good, document the migration steps. 

I just thought it was worth being transparent about what's on our plate going forward. 

Thanks!

fryz:
Released https://github.com/palantir/docker-compose-rule/releases/tag/0.34.0 which contains this.

vmykhailiuk:
[...]

PR #237, opened 2018-05-03
Fixed issue with memoized ports upon (re)start of docker container

mktange:
Port mappings can be different after restarting a docker container, i.e. when it's services uses ports in the ephemeral range. This means that the mapping needs to be re-evaluated upon (re)start of the container.


fryz:
Linking relevant issue: https://github.com/palantir/docker-compose-rule/issues/200

PR #241, opened 2018-06-14
Retrieve exit code for a container

### LINKED ISSUE #213: Get exit code of container
Hi there,

here a feature-request of mine.

Right now, it is already possible to get the [state](https://github.com/palantir/docker-compose-rule/blob/master/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/State.java#L19) of container.
It would be really nice, if the exit code (of a DOWN-container) could retrieved, too.

SchulteMarkus:
Fixes #213


SchulteMarkus:
Implementation notes:
I did not manage to get the docker-container-id for, in special, an container, which is down. Because of this, I decided to save the container-id when a container is being started. I am not to happy with this solution, but I did not find another way for being able to retrieve the exit code for a stopped/killed container.

SchulteMarkus:
Closed as PR starved.

PR #244, opened 2018-07-17

### LINKED ISSUE #172: logs are not reattached if container is restarted
if container in cluster fails to start but has restart: always in settings logs are saved only from first run of this container

pkoenig10:
Fixes #172 

This has the side effect of not collecting logs until the tests are finished running. But this seems acceptable given the use cases of this rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ public class DockerComposeRuleTest {
 }
 ```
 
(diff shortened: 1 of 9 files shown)


pkoenig10:
Closing in favor of #245

PR #245, opened 2018-07-17

pkoenig10:
Simpler alternative to #244 

This only saves the logs of the last run of the container. This is useful for containers that may be restarted multiple times before they are ready.


docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
```diff
@@ -100,7 +99,6 @@ public void parse_and_returns_container_names_on_ps() throws IOException, Interr
     @Test
     public void call_docker_compose_with_no_colour_flag_on_logs() throws IOException {
         when(executedProcess.getInputStream()).thenReturn(
-                toInputStream("id"),
                 toInputStream("docker-compose version 1.5.6, build 1ad8866"),
                 toInputStream("logs"));
```
dansanduleac: not your fault, but this feels super brittle :(

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -194,10 +195,6 @@ public String config() throws IOException, InterruptedException {
     @Override
     public boolean writeLogs(String container, OutputStream output) throws IOException {
         try {
-            Awaitility.await()
-                    .pollInterval(50, TimeUnit.MILLISECONDS)
-                    .atMost(LOG_WAIT_TIMEOUT.getMillis(), TimeUnit.MILLISECONDS)
-                    .until(() -> exists(container));
```
dansanduleac: why don't we wait for the container anymore?
If we later change the position of collectLogs then this might still break

dansanduleac: Ah so because we were waiting on all the containers, once they were all up (and we'd see the "docker-compose cluster started" log message), a container would go down in the interim, and then it'd time out here?

I think we should explicitly handle the case where a container _we_ started (and are tracking) went down, and fail fast in that case (while still collecting the logs).

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -200,7 +198,7 @@ public boolean writeLogs(String container, OutputStream output) throws IOExcepti
                     .until(() -> exists(container));
             Process executedProcess = followLogs(container);
             IOUtils.copy(executedProcess.getInputStream(), output);
-            executedProcess.waitFor(COMMAND_TIMEOUT.getMillis(), MILLISECONDS);
+            executedProcess.waitFor();
```
pkoenig10: Not sure why we had a timeout here. We probably shouldn't stop collecting logs after some arbitrary time if the container is still running. This process will be killed by the `shutdownNow` in `FileLogCollector#stopCollecting`.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/logging/FileLogCollector.java
```diff
@@ -28,14 +29,15 @@
...
+import org.joda.time.Duration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class FileLogCollector implements LogCollector {
 
     private static final Logger log = LoggerFactory.getLogger(FileLogCollector.class);
 
-    private static final long STOP_TIMEOUT_IN_MILLIS = 50;
+    private static final Duration STOP_TIMEOUT = standardSeconds(5);
```
pkoenig10: Needed to bump this up because 50 milliseconds is not enough time for the log collector to start collecting logs before it is shutdown if we call `DockerComposeRule#after` immediately after calling `DockerComposeRule#before`.

pkoenig10:
I couldn't figure out why the tests of the deprecated shutdown strategies were failing.

Given that these have been deprecated for [over a year](https://github.com/palantir/docker-compose-rule/commit/6ddadadaa14846e4cb86c4a8671463d7b483db68) it seemed reasonable to simply remove them.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
```diff
@@ -55,6 +38,8 @@
      */
     ShutdownStrategy SKIP = new SkipShutdownStrategy();
 
-    void shutdown(DockerCompose dockerCompose, Docker docker) throws IOException, InterruptedException;
+    void stop(DockerCompose dockerCompose) throws IOException, InterruptedException;
+
+    void shutdown(DockerCompose dockerCompose) throws IOException, InterruptedException;
```
iamdanfox: I think if you're going to turn a one-method interface into a two method interface then it's really worth providing some javadoc to describe the semantics of these... especially as `shutdown` is a DCR concept, not a vanilla docker one

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -148,19 +148,23 @@ public void before() throws IOException, InterruptedException {
...
         upDockerCompose.up();
 
-        logCollector().startCollecting(dockerCompose());
         log.debug("Waiting for services");
         new ClusterWait(ClusterHealthCheck.nativeHealthChecks(), nativeServiceHealthCheckTimeout())
                 .waitUntilReady(containers());
         clusterWaits().forEach(clusterWait -> clusterWait.waitUntilReady(containers()));
         log.debug("docker-compose cluster started");
+
+        logCollector().startCollecting(dockerCompose());
```
iamdanfox: What if a docker container fails to start up? Won't the log collector not collect _anything_ for that container? This would be super hard to debug :(

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/KillDownShutdownStrategyShould.java
```diff
@@ -14,14 +14,23 @@
 public class KillDownShutdownStrategyShould {
 
     @Test
-    public void call_kill_then_down() throws Exception {
+    public void stop_call_kill() throws Exception {
```
iamdanfox: The pattern elsewhere in this repo is that Class#method should read like a vaguely english sentence:

```
KillDownShutdownStrategyShould#call_kill_then_down (kinda expanatory)
KillDownShutdownStrategyShould#stop_call_kill (not quite as clear)
```

I'd propose naming these something like `kill_container_when_stop_method_is_called`

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/GracefulShutdownStrategy.java
```diff
@@ -18,11 +18,16 @@
...
 
     @Override
-    public void shutdown(DockerCompose dockerCompose, Docker docker) throws IOException, InterruptedException {
-        log.debug("Killing docker-compose cluster");
-        dockerCompose.down();
+    public void stop(DockerCompose dockerCompose) throws IOException, InterruptedException {
+        log.debug("Stopping docker-compose cluster");
+        dockerCompose.stop();
         dockerCompose.kill();
-        dockerCompose.rm();
```
iamdanfox: Was the `.rm()` intentionally deleted?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
iamdanfox: I think the rule of thumb was that every single `.waitFor` should have some kind of timeout to be 100% certain that tests will never hang forever

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
pkoenig10: The concepts here align with the docker concepts of `stop` and `rm`. docker-compose `down` is just `stop` + `rm`. I'll update the method names to make this more clear.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
pkoenig10: Yeah this a downside, but I'm not sure how it can be mitigated. The issue is that you can't follow logs across container restarts.

An alternative is that we don't do any log collection until the end (this was my original implementation in #244). But that can cause the log collection to take a long time if your containers have generated a lot of logs. And we lose the benefit of having logs while the container is still running.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/GracefulShutdownStrategy.java
pkoenig10: It is redundant, `.down()` implies `.rm()`.

pkoenig10: The act of removing containers has been separated in the `shutdown` method.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
pkoenig10: Looking at this again, it's not clear to me why we need this `waitFor` at all. This is the last things we do in a executor thread so there's really no need to wait here.

pkoenig10:
This PR has been updated to collect logs at the end. This allows us to collect logs across restarts and collect logs if we fail to pass health checks. The downside is that we don't collect logs live while the container is running, but this seems like the correct tradeoff for the intended use case of this library.

gregakinman:
should probably remove this test then? https://github.com/palantir/docker-compose-rule/pull/245/files#diff-e2b2cc50521a94cd991b1a3faaadc5e3R230

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
```diff
@@ -1,63 +0,0 @@
...
-import org.slf4j.LoggerFactory;
-
-/**
- * Shuts down containers as fast as possible, without giving them time to finish
- * IO or clean up any resources.
- *
- * @deprecated Use {@link ShutdownStrategy#KILL_DOWN}
- */
-@Deprecated
-public class AggressiveShutdownStrategy implements ShutdownStrategy {
```
dansanduleac: Why are we removing deprecated classes in this PR?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
```diff
@@ -47,14 +29,16 @@
      */
     ShutdownStrategy KILL_DOWN = new KillDownShutdownStrategy();
     /**
-     * Skip shutdown, leaving containers running after tests finish executing.
+     * Skip down, leaving containers running after tests finish executing.
```
dansanduleac: intentional?

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/SkipShutdownStrategyShould.java
```diff
@@ -0,0 +1,32 @@
...
+
+package com.palantir.docker.compose.execution;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verifyZeroInteractions;
+
+import com.palantir.docker.compose.configuration.ShutdownStrategy;
+import org.junit.Test;
+
+public class SkipShutdownStrategyShould {
```
dansanduleac: convention is for test classes to end in Test

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
pkoenig10: Yes, "down" is a docker-compose term and is more clear than "shutdown" which is not a docker-compose term.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
pkoenig10: I had trouble getting the tests to pass for these shutdown strategies. Given that this PR changes the behavior of the shutdown strategies, it seemed like an appropriate time to remove them.

pnepywoda:
@dansanduleac what else needs to be done here to get it merged?

pkoenig10:
Closing in favor of #293

PR #249, opened 2018-09-04
Adding container name to error message along with internal port, help…

saavkaar:
Adding container name to error message along with internal port, helpful when you have multiple containers with service running on same internal port.


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Container.java
```diff
@@ -60,7 +60,7 @@ public SuccessOrFailure portIsListeningOnHttp(int internalPort, Function<DockerP
                 return SuccessOrFailure.failure("Internal port " + internalPort + " is not listening in container " + containerName);
             }
             return port.isHttpRespondingSuccessfully(urlFunction, andCheckStatus)
-                    .mapFailure(failureMessage -> internalPort + " does not have a http response from " + urlFunction.apply(port) + ":\n" + failureMessage);
+                    .mapFailure(failureMessage -> containerName+":"+internalPort + " does not have a http response from " + urlFunction.apply(port) + ":\n" + failureMessage);
```
bavardage:
```suggestion
                    .mapFailure(failureMessage -> containerName + ":" + internalPort + " does not have a http response from " + urlFunction.apply(port) + ":\n" + failureMessage);
```

PR #252, opened 2018-09-21
Fixes #250 'Throwables other than Exceptions are not propagated to the main thread '

### LINKED ISSUE #250: Throwables other than Exceptions are not propagated to the main thread
# Type: bug ;; Repro rate: 100%

# Repro steps:
1. Create a `HealthCheck` which throws some descendant of `Throwable` which is not an `Exception` (e.g. an `Error` - `NoSuchMethodError`)
2. Use this `HealthCheck` in a `DockerComposeRule`.

## Actual result:
The rule fails to start with `ConditionTimeoutException`.

## Expected result:
The rule fails to start with the actual `Throwable` thrown bt the `HealthCheck`.

# Severity
My `HealthCheck` was throwing a `NoSuchMethodError`. It took me days to find out the actual issue behind this was. It was very hard to diagnose as the failure is quite misleading (`ConditionTimeoutException`) - and when you look at it it actually looks like the code hangs on the line which causes the `NoSuchMethodError` (also from the IDE).

# Cause
`DockerComposeRule` uses Awaitility to wait for `HealthCheck`s to resolve. Awaitility in turn uses a `ScheduledExecutorService` which has an infamous feature of swallowing exceptions thrown in the jobs submitted for execution. Awaitility does some work to propagate the errors to the main thread but as it turns out it does it only for `Exception` descendants, excluding other `Throwables` (e.g. `IncompatibleClassChangeError` inheritors which are caused by runtime dependency problems and thus are quite common - `NoSuchMethodError`, `NoSuchFieldError`, etc.).

A simple test for this:

    @Test
    public void shouldNotTimeOut() {
        assertThatThrownBy(() ->
                Awaitility
                        .await()
                        .pollInterval(50, TimeUnit.MILLISECONDS)
                        .atMost(10, TimeUnit.SECONDS)
                        .until(() -> {
                            throw new AnError();
                        }))
                .isNotInstanceOf(ConditionTimeoutException.class);
    }

# Solution
Upgrade Awaitility to 1.7.0 which copes a bit better with this (it throws an `ExecutionException` in the case discussed). Or do what we chose to do in our code - wrap the HealthCheck and propagate the fault as an Exception:

    public static <T> HealthCheck<T> fixHealthCheck(HealthCheck<T> delegate) {
        return container -> {
            try {
                return delegate.isHealthy(container);
            } catch (Throwable e) {
                throw new RuntimeException(
                        "Unexpected exception occurred while performing health check.",
                        e);
            }
        };
    }

jannis-baratheon:
This PR fixes #250


palantirtech:
Thanks for your interest in [palantir/docker-compose-rule](https://github.com/palantir/docker-compose-rule), @jannis-baratheon! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

jannis-baratheon:
Could anyone take a look?

PR #271, opened 2019-06-21
add shutdown strategy which calls a callback before shutting down

bavardage:
this can be useful for gathering extra information from containers (copying files out, etc)

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/CallbackThenDelegateShutdownStrategy.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/CallbackThenDelegateShutdownStrategy.java
@@ -35,7 +35,10 @@ public class CallbackThenDelegateShutdownStrategy implements ShutdownStrategy {
 
     @Override
     public void shutdown(DockerCompose dockerCompose, Docker docker) throws IOException, InterruptedException {
-        callback.run();
-        delegate.shutdown(dockerCompose, docker);
+        try {
+            callback.run();
+        } finally {
+            delegate.shutdown(dockerCompose, docker);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


bavardage:
blocked on #272

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/CallbackThenDelegateShutdownStrategy.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public CallbackThenDelegateShutdownStrategy(ShutdownStrategy delegate, Runnable callback) {
+        this.delegate = delegate;
+        this.callback = callback;
+    }
+
+    @Override
+    public void shutdown(DockerCompose dockerCompose, Docker docker) throws IOException, InterruptedException {
+        callback.run();
+        delegate.shutdown(dockerCompose, docker);
```
CRogers: Might be a good idea to run this as a finally just in case `callback.run()` throws.

bavardage: yeah good point - updated in e1a736e27b71b213f811c7c4aae21f7a09080329

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/CallbackThenDelegateShutdownStrategyShould.java
```diff
@@ -40,4 +45,29 @@ public void call_callback_then_call_delegate() throws Exception {
...
+
+        RuntimeException callbackException = new RuntimeException("exception in callback");
+        doThrow(callbackException).when(callback).run();
+
+        DockerCompose dockerCompose = mock(DockerCompose.class);
+        Docker docker = mock(Docker.class);
+
+        try {
+            ShutdownStrategy.callbackAndThen(callback, delegate).shutdown(dockerCompose, docker);
+            fail("expected exception");
```
bavardage: hm, would have been nicer to use assertj's `assertThatThrownBy`, but didn't want to move all the tests over to using that over hamcrest `assertThat`

CRogers: I think it's acceptable to have both on the classpath and use them in different test classes. No need to write more hamcrest if the we can use the more discoverable/readable assertj.

PR #288, opened 2019-06-27
[break] Stats on docker compose rule runs

CRogers:
## Before this PR
There's no way to inspect how long it takes to pull images, wait for containers to spin up, shutdown containers etc.

## After this PR
==COMMIT_MSG==
Stats on how long various operations take are recorded by docker compose rule, and an extension point is provided to get these statistics.
==COMMIT_MSG==

Internally, we want to instrument docker compose rule to see how long various parts of it take to run/where people are experiencing pain with long running ETEs. This will allow use to record this information and take appropriate action on the data.

## Possible downsides?
I have made this no longer an `ExternalResource`, as there is a long standing DCR problem where `after()` is not run if `before()` fails, meaning your docker containers/networks still exist even when the test failed. Now they will be wiped up (can use the `ShutdownStrategy` option to change this). Better `after()` behaviour is required so we can actually report stats even when `before()` has failed. Technically, this could be a break if users do the following:

```
ExternalResource dockerComposeRule = DockerComposeRule.builder()....build();
```

but I honestly think that would be a very strange thing to do.


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/CustomImmutablesStyle.java
```diff
@@ -20,5 +20,5 @@
 import org.immutables.value.Value.Style;
 
 @Target({ElementType.PACKAGE, ElementType.TYPE})
-@Style(depluralize = true, strictBuilder = true)
+@Style(depluralize = true, strictBuilder = true, overshadowImplementation = true)
```
CRogers: This is also technically a break as people may have done `ImmutableDockerComposeRule dcr = DockerComposeRule.builder()....build()` in the past, which is now a `DockerComposeRule`. A few people have done this: https://github.com/search?q=ImmutableDockerComposeRule&type=Code, and 1 further case internally, but that's it. I need the later behaviour in order to instrument the DCR class properly.

docker-compose-rule-core/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,54 @@
...
+
+        <Logger name="com.palantir.dia" level="debug" additivity="false">
+            <AppenderRef ref="STDOUT"/>
+            <AppenderRef ref="LOG"/>
+        </Logger>
+
+        <Logger name="org.apache.camel" level="debug" additivity="false">
+            <AppenderRef ref="STDOUT"/>
+            <AppenderRef ref="LOG"/>
+        </Logger>
```
carterkozak: We can remove the `com.palantir.dia` and `org.apache.camel` blocks, remove appenders and `additivity` from the `com.palantir` block in order to get the same output with a more concise configuration.

CRogers: I just decided to split the logging changes into another PR: https://github.com/palantir/docker-compose-rule/pull/289

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterWait.java
```diff
@@ -30,10 +31,31 @@
     private static final Logger log = LoggerFactory.getLogger(ClusterWait.class);
     private final ClusterHealthCheck clusterHealthCheck;
     private final ReadableDuration timeout;
+    private final Listener listener;
+
+    public interface Listener {
+        void becameHealthy();
+        void timedOut();
```
iamdanfox: default methods?

CRogers:
Closing this PR as we're going to try for a different method, emitting events from DCR that we can receive elsewhere to construct stats, which should be easier to avoid ABI breaks and make it easier to control what we measure in the future!

PR #289, opened 2019-06-27
Fix logging in tests

CRogers:
## Before this PR
No logging output appears in the tests. This error appears

```
log4j:WARN No appenders could be found for logger (com.palantir.docker.compose.execution.DockerComposeExecutable).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

## After this PR
Logging output appears in the test, by using slf4j-simple rather than complicated log4j that has become misconfigured


dansanduleac:
👍

PR #292, opened 2019-07-01
[improvement] Emit events to be used for stats/metrics tracking

CRogers:
## Before this PR
It is very hard/impossible to introspect how long different parts of docker compose rule takes to run, why they fail etc.

## After this PR
==COMMIT_MSG==
Events are emitted for pull, build, up, waiting for services and shutdown.
==COMMIT_MSG==

These events allow users to time how long parts of DCR take and whether they work. In particular, timing of per container events is now possible.

## Possible downsides?
Increased API surface that we have to support.


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/waiting/ClusterWaitInterface.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.docker.compose.connection.waiting;
+
+import com.palantir.docker.compose.connection.Cluster;
+
+public interface ClusterWaitInterface {
```
CRogers: I had to extract out this interface, however `ClusterWait` needs to remain the same. Open to a better name.

dansanduleac: why did we have to extract this? it doesn't look like there are other implementors except `ClusterWait`

CRogers: There's a sneaky one here: https://github.com/palantir/docker-compose-rule/pull/292/files#diff-fd69676598fa179a6d84c0bd468b2eeeR85

CRogers: Actually, you're right, that doesn't even need to be an interface any more. At one put I was adding these things in the builder, now that's no longer happening it should be fine.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private void emitNotThrowing(CheckedRunnable runnable, TaskEvent.Factory factory) {
+        try {
+            emitThrowing(runnable, factory);
+        } catch (InterruptedException | IOException e) {
+            Throwables.propagate(e);
+        }
+    }
+
+    private void emitThrowing(CheckedRunnable runnable, TaskEvent.Factory factory)
```
iamdanfox: wrap

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
```diff
@@ -0,0 +1,134 @@
...
+
+        return cluster -> emitNotThrowing(
+                () -> recordingClusterWait.accept(cluster),
+                ClusterWaitEvent.factory(
+                        () -> recordedServiceNames.get().orElseThrow(
+                                () -> new IllegalStateException("Recorded service names have not yet been computed")),
+                        clusterWaitType));
+    }
+
+    private void emitNotThrowing(CheckedRunnable runnable, TaskEvent.Factory factory) {
```
iamdanfox: wrapUnchecked

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/events/TaskEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.docker.compose.events;
+
+public interface TaskEvent extends DockerComposeRuleEvent {
```
iamdanfox: javadoc to say this is a marker interface?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -133,34 +152,78 @@ protected LogCollector logCollector() {
...
-    @Override
+    private void waitForAllClusterWaits(List<Consumer<Cluster>> allClusterWaits) throws InterruptedException {
+        ListeningExecutorService executorService = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(
+                allClusterWaits.size(),
+                new ThreadFactoryBuilder()
+                        .setNameFormat("dcr-wait-%d")
+                        .build()));
+
+        try {
+            ListenableFuture<List<Object>> listListenableFuture =
```
iamdanfox: `<?>` as we don't use it?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public Consumer<Cluster> userClusterWait(ClusterWait clusterWait)  {
+        return clusterWait(ClusterWaitType.USER, clusterWait);
+    }
+
+    public Consumer<Cluster> nativeClusterWait(ClusterWait clusterWait) {
+        return clusterWait(ClusterWaitType.NATIVE, clusterWait);
+    }
+
+    private Consumer<Cluster> clusterWait(
```
iamdanfox: gotta justify why we have to do this - clusterwaits don't declare which containers they access, so we have to figure it out while they are executing??

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/events/DockerComposeRuleEvent.java
```diff
@@ -0,0 +1,25 @@
...
+
+package com.palantir.docker.compose.events;
+
+import org.immutables.value.Value;
+
+@Value.Style(
+        visibility = Value.Style.ImplementationVisibility.PACKAGE,
+        builderVisibility = Value.Style.BuilderVisibility.PACKAGE
+)
+public interface DockerComposeRuleEvent { }
```
iamdanfox: Maybe a little javadoc about why we're adding this and why people might care?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/events/ClusterWaitEvent.java
```diff
@@ -0,0 +1,97 @@
...
+import com.palantir.docker.compose.events.TaskEvent.Succeeded;
+import java.util.Set;
+import java.util.function.Supplier;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.immutables.value.Value;
+
+public interface ClusterWaitEvent extends DockerComposeRuleEvent {
+
+    ClusterWaitEventId eventId();
+    ClusterWaitType clusterWaitType();
```
iamdanfox: :+1: big fan of nice types :)

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
CRogers: I think this is tricky as we call out `emit`, which uses the a field in this class. Unless we were to pass the `emit` function to `wrap`

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
```diff
@@ -0,0 +1,162 @@
...
+            emitEvent(event);
+        }
+    }
+
+    private void emitEvent(Event event) {
+        eventConsumers.forEach(eventConsumer -> {
+            try {
+                eventConsumer.receiveEvent(event);
+            } catch (Exception e) {
+                log.error("Error sending event {}", event, e);
```
ferozco: I think we should propagate the error if an eventConsumer fails

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/events/EventConsumer.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.docker.compose.events;
+
+public interface EventConsumer {
```
ferozco: could we use `Consumer<Event>` instead?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/RecordingCluster.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.common.collect.Sets;
+import com.palantir.docker.compose.connection.Cluster;
+import com.palantir.docker.compose.connection.Container;
+import com.palantir.docker.compose.connection.ContainerCache;
+import java.io.IOException;
+import java.util.List;
+import java.util.Set;
+
+final class RecordingCluster extends Cluster {
+    private final Cluster realCluster;
```
ferozco: delegate?

docker-compose-rule-junit4/src/integrationTest/java/com/palantir/docker/compose/EventsIntegrationTest.java
```diff
@@ -0,0 +1,169 @@
...
+                .file(ALL_GOOD_DOCKER_COMPOSE_YAML)
+                .waitingForService(
+                        "one",
+                        container -> SuccessOrFailure.failure(failureMessage),
+                        org.joda.time.Duration.standardSeconds(1))
+                .addEventConsumer(eventConsumer)
+                .build();
+
+        try {
+            runDockerComposeRule(dockerComposeRule);
```
ferozco: you probably want to ensure that an exception is thrown

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -133,34 +152,78 @@ protected LogCollector logCollector() {
...
+
+        waitForAllClusterWaits(allClusterWaits);
+
         log.debug("docker-compose cluster started");
     }
 
-    @Override
+    private void waitForAllClusterWaits(List<Consumer<Cluster>> allClusterWaits) throws InterruptedException {
+        ListeningExecutorService executorService = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(
+                allClusterWaits.size(),
```
ferozco: do we need a thread per consumer?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -133,34 +152,78 @@ protected LogCollector logCollector() {
...
+                new ThreadFactoryBuilder()
+                        .setNameFormat("dcr-wait-%d")
+                        .build()));
+
+        try {
+            ListenableFuture<?> listListenableFuture =
```
ferozco: why does this need to be done in parallel?

docker-compose-rule-events-api/src/main/conjure/docker-compose-rule-event-api.yml
```diff
@@ -0,0 +1,49 @@
...
+      ClusterWaitType:
+        values:
+            - NATIVE
+            - USER
+
+      ShutdownEvent:
+        fields:
+          task: Task
+
+      Task:
```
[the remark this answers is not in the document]
CRogers: I originally had it like this, but couldn't think of a use case that warranted the complexity/all the extra events. For our use case we want work out how long it took, so breaking the task up into smaller events then having to join them back together in our <internal analysis platform> seems a lot of effort. We're also not precluded from making start/stop events in the future if someone needs them (in fact this current event would kind of end up as a [Summary Event](http://verraes.net/2019/05/patterns-for-decoupling-distsys-summary-event/))

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
CRogers: What happens if you have multiple event consumers? Then if one fails to consume an event the rest don't get the event.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
CRogers: Yep, we need one thread per cluster wait. Since we only know when a service becomes healthy by checking, if we check them serially a service that becomes healthy quickly could seem like it takes forever as it's checked after a slow starting service has become healthy.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
CRogers: Ditto for measuring how long it takes for each service to become healthy.

docker-compose-rule-events-api/src/main/conjure/docker-compose-rule-event-api.yml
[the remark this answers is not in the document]
ferozco: Appreciate the link :)

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
ferozco: I'm trying to avoid unexpected behaviour when a user provides their own consumers. Its kinda weird, but I could imagine users adding event consumers to ensure some sort of preconditions, and failing the test (suite|case) if they do not hold

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
CRogers: Welp, I tested this on an internal product JT that has 9 services in it's docker compose and it actually fails to startup due to increased CPU usage using this version of DCR, probably from all polling. I think we need to do something like use a higher pollInterval (maybe it becomes every 1 sec after 2 secs of trying?)  or a `ScheduledExecutorService` instead to get it back below the timeout.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
```diff
@@ -0,0 +1,162 @@
...
+    private Consumer<Cluster> clusterWait(
+            ClusterWaitType clusterWaitType,
+            ClusterWait clusterWait) {
+        // This weird bit of complexity is because we can't tell what services a ClusterWait is using until it
+        // actually runs. So we have to record the services it accesses then use this to generate the events. The
+        // Optional exists solely as a check again logic errors - in the case when events are generated before the
+        // cluster wait has begun.
+        AtomicReference<Optional<Set<String>>> recordedServiceNames = new AtomicReference<>(Optional.empty());
+
+        Consumer<Cluster> recordingClusterWait = cluster -> {
```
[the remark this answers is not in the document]
CRogers: Moved it out it's own class instead

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/events/EventConsumer.java
CRogers: I actually like to keep this a concrete type for the API. It makes it easier to search for people using it on github, if we want to add more defaulted methods etc onto it in the future we can.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
CRogers: Now poll every 1 second to reduce CPU load.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/EventEmitter.java
CRogers: Ok, this will now throw and add all exceptions that happened as suppressed.

PR #293, opened 2019-07-01

### LINKED ISSUE #172: logs are not reattached if container is restarted

pkoenig10:
Fixes #172

## Before this PR
Logs of restarted containers are not collected.

## After this PR
Logs of restarted containers are collected.

This required splitting the shutdown strategy into two steps, stop and kill.


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerCompose.java
```diff
@@ -44,6 +45,6 @@ static Version version() throws IOException, InterruptedException {
     Optional<String> id(Container container) throws IOException, InterruptedException;
     String config() throws IOException, InterruptedException;
     List<String> services() throws IOException, InterruptedException;
-    boolean writeLogs(String container, OutputStream output) throws IOException;
+    void writeLogs(String container, OutputStream output) throws IOException, InterruptedException;
```
CRogers: I think this is technically a breaking change - can we return a boolean still and not propagate the interrupted exception?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
```diff
@@ -25,42 +22,14 @@
     private static final Logger log = LoggerFactory.getLogger(AggressiveShutdownStrategy.class);
 
     @Override
-    public void shutdown(DockerCompose dockerCompose, Docker docker) throws IOException, InterruptedException {
-        List<ContainerName> runningContainers = dockerCompose.ps();
-
-        log.info("Shutting down {}", runningContainers.stream().map(ContainerName::semanticName).collect(toList()));
-        if (removeContainersCatchingErrors(docker, runningContainers)) {
```
CRogers: This seems like a lot of effort to just do `docker.kill()` - I wonder if there was a good reason behind it?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
```diff
@@ -60,6 +59,8 @@ static ShutdownStrategy callbackAndThen(Runnable callback, ShutdownStrategy shut
         return new CallbackThenDelegateShutdownStrategy(shutdownStrategy, callback);
     }
 
-    void shutdown(DockerCompose dockerCompose, Docker docker) throws IOException, InterruptedException;
```
CRogers: Removing this is a break as `ShutdownStrategy` is public api. Can we keep `shutdown(DockerCompose, Docker)` as a default method (and deprecate it), then call out to `stop` and `down` which are default empty methods (but implemented by everything in the main library?)? That way if anyone has implemented their own shutdown strategy they will not be broken?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
pkoenig10: I think the intention was to kill each container individually to do a best-effort shutdown. But given that docker-compose-rule is used almost exclusively for CI builds, this seems unnecessary.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
CRogers: Actually, that might not work as we can't just call shutdown, we want to call `stop` then `down`. Maybe keeping an empty deprecated default `shutdown(...)` method and calling it from `stop(...)` would be the actual behaviour we want?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
CRogers: Also, I think `docker-compose kill` literally does this.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -159,8 +158,11 @@ public void before() throws IOException, InterruptedException {
     @Override
     public void after() {
         try {
-            shutdownStrategy().shutdown(this.dockerCompose(), this.docker());
-            logCollector().stopCollecting();
+            shutdownStrategy().stop(this.dockerCompose());
+
+            logCollector().collectLogs(this.dockerCompose());
```
CRogers: Does this mean it just collects logs once at shutdown? As in it no longer streams logs?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/AggressiveShutdownStrategy.java
pkoenig10: I think you're right

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
pkoenig10: Correct, this is an explicit tradeoff. See:
- https://github.com/palantir/docker-compose-rule/pull/245#discussion_r215345662
- https://github.com/palantir/docker-compose-rule/pull/245#issuecomment-424485444

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
pkoenig10: The current implementation of `shutdown` is effectively just `docker-compose down` (with some implementations calling `docker-compose stop` first to give processes a chance to end cleanly). I think it's better to preserve the existing behavior with the downside that you don't get logs if you call `shutdown` manually (but you really shouldn't be doing that anyways).

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
CRogers: Ah, I didn't see the previous PR.

I guess if the problem is the `docker-compose logs` process is dying, you could make an `InputStream` that transparently waits until the container is started again, then run `docker-compose -f --tail 0` to get the rest of the logs, maybe using `--timestamps` and some other number of rows to `--tail` to ensure you get them all. What do you think?

Then again, that seems quite fickle and it is true that I for example pretty much always just run `dc logs -f` locally.

CRogers: Also, on another very important note, if cluster fails to start up in `before()`, `after()` will not be run (see `ExternalResource` and how bad it is). I think this is indicative that we should add an integTest to check that logs are collected if the cluster fails to start up, and for this PR when the container is restarted, just to make sure we maintain these behaviours in the future.

To fix the `ExternalResource` problem, the option is to stop using `ExternalResource` and write our own apply method that puts the `before` call in the `try` block. This is probably the behaviour we always wanted anyway.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
CRogers: The thing is I don't think people will be calling `shutdown` manually - it's DCR itself executing the `ShutdownStrategy`, so after this change if someone had written a custom `ShutdownStrategy` and gives it to DCR nothing will happen. Hence why we need to make `stop()` or `down()` call `shutdown()`, so the user's old shutdown code still works.

pkoenig10: If a user wrote a custom `ShutdownStrategy` it would no longer compile because they don't implement the `stop()` and `down()` methods.

CRogers: That's what I'm saying we should try to avoid - people should not experience a break from this change, if possible. They can be made default methods to not cause breaks.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
pkoenig10: That seems incredibly complicated. Dealing with restarting containers is very difficult and we're not really giving up anything by collecting logs at the end. I would strongly prefer to keep this implementation simple and have something that work for the 99.9% of normal cases.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/configuration/ShutdownStrategy.java
pkoenig10: Changing `down()` to call `shutdown()` requires also passing in `Docker` which I'd like to avoid (since you shouldn't need to use docker directly to shut down). Instead let's just continue to call `shutdown()` directly and make it a no-op default implementation.

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
CRogers: Fair enough

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -38,13 +38,15 @@
...
-import org.junit.rules.ExternalResource;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @Value.Immutable
 @CustomImmutablesStyle
-public abstract class DockerComposeRule extends ExternalResource {
```
CRogers: Super technically this is a break, but I'm planning on making a (small) actual break tomorrow, so maybe once we merge this don't release and I'll group them together into a major version bump release at the same time?

pkoenig10: Sounds good to me

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
```diff
@@ -98,10 +98,10 @@ public void parse_and_returns_container_names_on_ps() throws IOException, Interr
     }
 
     @Test
-    public void call_docker_compose_with_no_colour_flag_on_logs() throws IOException {
+    public void call_docker_compose_with_no_colour_flag_on_logs() throws IOException, InterruptedException {
         when(executedProcess.getInputStream()).thenReturn(
-                toInputStream("id"),
-                toInputStream("docker-compose version 1.5.6, build 1ad8866"),
+//                toInputStream("id"),
```
CRogers: ?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
pkoenig10: The methods in `ExternalResource` are non-final, so we could avoid a break here by keeping the extends and overriding the methods

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
pkoenig10: Fixed

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
pkoenig10: Done

CRogers:
Just needs changelog

pkoenig10:
@CRogers I don't have permission to edit the changelog-app comment. Could you do it?

CRogers:
Next time it's probably best to push your branches directly to this repo rather than your own fork, it makes the permissions all funny etc (and harder for me to make changes on your branch)

pkoenig10:
@CRogers added a changelog, any other blockers to merging this?

PR #299, opened 2019-07-11
Increase timeout and simplify log collection

pkoenig10:
## Before this PR
#293 broke log collection because the we time out the log collection after 50ms. This made sense when we were streaming logs but not after we changed the behavior to collect at the end.

## After this PR
We now give the logs process 1 minute to collect logs before timing out.

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
@@ -195,14 +195,15 @@ public class DefaultDockerCompose implements DockerCompose {
         try {
             Process executedProcess = logs(container);
             IOUtils.copy(executedProcess.getInputStream(), output);
-            boolean processExecuted = executedProcess.waitFor(LOG_TIMEOUT.getMillis(), TimeUnit.MILLISECONDS);
-            if (!processExecuted) {
+            boolean processFinished = executedProcess.waitFor(LOG_TIMEOUT.getMillis(), TimeUnit.MILLISECONDS);
+            boolean timedOut = !processFinished;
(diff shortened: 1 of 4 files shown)


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -194,17 +195,12 @@ public boolean writeLogs(String container, OutputStream output) throws IOExcepti
         try {
             Process executedProcess = logs(container);
             IOUtils.copy(executedProcess.getInputStream(), output);
-            executedProcess.waitFor();
-            return true;
+            return executedProcess.waitFor(LOG_TIMEOUT.getMillis(), TimeUnit.MILLISECONDS);
```
CRogers: We probably want to `destroyForcibly` the process if it times out - I think `waitFor` keeps it running.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/logging/FileLogCollector.java
```diff
@@ -55,40 +47,31 @@ public static LogCollector fromPath(String path) {
...
     public void collectLogs(DockerCompose dockerCompose) throws IOException, InterruptedException {
-        if (executor != null) {
-            throw new RuntimeException("Cannot start collecting the same logs twice");
-        }
-
-        List<String> serviceNames = dockerCompose.services();
-        if (serviceNames.size() == 0) {
-            return;
-        }
-        executor = Executors.newFixedThreadPool(serviceNames.size());
```
CRogers: Yeah, there really is no need for all this concurrency now.

CRogers:
:+1:

PR #303, opened 2019-07-16
Refactor to centralize implementation in docker-compose-rule-core

ferozco:
## Before this PR
Although the majority of interesting logic lived within `docker-compose-rule-core` there was still a bit of setup within `docker-compose-rule-junit4` which made it difficult to create a junit-jupiter extension.

## After this PR
Refactor all logic from `DockerComposeRule` to a test runner agnostic `DockerComposeManager` which `DockerComposeRule` inherits from.

## Possible downsides?
Since `DockerComposeRule` no longer extends `org.junit.rules.ExternalResource` and instead implements `org.junit.rules.TestRule`, usage DCR with junit-vintage will no longer work.


docker-compose-rule-core/src/test/java/com/palantir/docker/compose/EventEmitterShould.java
```diff
@@ -81,7 +80,7 @@ public void produce_started_then_failed_for_failing_build_run() {
         OffsetDateTime startedTime = timeIs(5);
         AtomicReference<OffsetDateTime> endTime = new AtomicReference<>();
 
-        assertThatExceptionOfType(IllegalStateException.class).isThrownBy(() ->
+        Assertions.assertThatExceptionOfType(IllegalStateException.class).isThrownBy(() ->
```
CRogers: Why `Assertions.`?

docker-compose-rule-core/src/integrationTest/java/com/palantir/docker/compose/DockerComposeManagerUpContainerIntegrationTest.java
```diff
@@ -27,38 +27,41 @@
...
-    public final DockerComposeRule dockerComposeRule = DockerComposeRule
-            .builder()
-            .shutdownStrategy(AGGRESSIVE)
-            .file("src/test/resources/up-integration-test.yaml")
-            .build();
+    private DockerComposeManager dockerComposeManager;
+
+    @Before
+    public void before() {
+        dockerComposeManager = DockerComposeManager.testBuilder()
```
CRogers: Shouldn't this be assigned at the field and we should call `before()` here?

docker-compose-rule-core/src/integrationTest/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
```diff
@@ -29,20 +29,20 @@
 
     @Test
     public void shut_down_multiple_containers_immediately() throws Exception {
-        DockerComposeRule rule = DockerComposeRule.builder()
+        DockerComposeManager docker = DockerComposeManager.testBuilder()
```
dansanduleac: shall we move all these tests into docker-compose-rule-core, since they're not using the rule anymore but only the manager?

docker-compose-rule-core/src/integrationTest/java/com/palantir/docker/compose/DockerComposeManagerNativeHealthcheckIntegrationTest.java
```diff
@@ -59,32 +59,32 @@ public void shutdownPool() {
...
 
-        Container container = rule.containers().container("dummy");
+        Container container = docker.containers().container("withHealthcheck");
         await().until(container::state, Matchers.equalTo(State.UNHEALTHY));
 
-        // The "dummy" container should not initially pass its healthcheck
+        // The "withHealthCheck" container should not initially pass its healthcheck
         try {
             getUninterruptibly(beforeFuture, 1, TimeUnit.SECONDS);
             fail("Expected before() to wait");
```
dansanduleac: There's an errorprone rule out there to migrate this insanity to a proper assertj I think!
Should consider doing that afterwards as it makes it much easier to read

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/DockerComposeManager.java
```diff
@@ -0,0 +1,380 @@
...
+
+    public String run(DockerComposeRunOption options, String containerName,
+            DockerComposeRunArgument arguments) throws IOException, InterruptedException {
+        return dockerCompose().run(options, containerName, arguments);
+    }
+
+    /*
+     *  TODO(forozco)
+     */
+    public static Builder testBuilder() {
```
CRogers: Why is this called `testBuilder` rather than `builder`?

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
```diff
@@ -1,160 +1,42 @@
...
-    protected LogCollector logCollector() {
-        return new DoNothingLogCollector();
-    }
-
-    @Value.Derived
-    protected EventEmitter emitEventsFor() {
-        return new EventEmitter(eventConsumers());
-    }
-
+public abstract class DockerComposeRule extends DockerComposeManager implements TestRule {
```
CRogers: It's kind of sad we can't keep compat with `ExternalResource` here for those who have used the vintage runner. The reason we can't do it is because of inheritance - can we use delegation here instead? A benefit is that we could make the junit4/junit5 versions exposed as interfaces instead of classes, which would mean we can avoid using Javassist to make `EnhancedDockerComposeRule` as well. Also, subclassing immutables classes has always seemed a little weird to me. It also means we could make DockerComposeManager package private and reduce our API surface?

docker-compose-rule-core/src/integrationTest/java/com/palantir/docker/compose/AggressiveShutdownStrategyIntegrationTest.java
ferozco: They are in docker-compose-rule core!

docker-compose-rule-core/src/integrationTest/java/com/palantir/docker/compose/DockerComposeManagerNativeHealthcheckIntegrationTest.java
ferozco: Happy to do in a follow up

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/DockerComposeManager.java
ferozco: Sorry, I meant to flesh out the comment. To get `DockerComposeRule` and `DockerComposeExtension` working with as little duplicate code as possible, I had to make them both extend `DockerComposeManager`. Unfortunately, its not possible to get Immutable builders to extend from each other. So to get `DockerComposeRule.builder()` to worker we could not have a `builder()` method on `DockerComposeManager`

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
dansanduleac: Caught up with @ferozco over zoom, think we can make DockerComposeMananger private-ish anyway by moving it to another project which is not part of the api exposed by `docker-compose-rule-(junit4|jupiter)`.
WRT people who have migrated to using this via the vintage runner, there shouldn't be many cases of this, and we are planning to cut a major version anyway to make it super clear that things are a bit different.

It would be nice-ish to make these interfaces, but that also means we'd have to write a bunch of delegation code twice which is also not pretty :(

docker-compose-rule-core/src/integrationTest/java/com/palantir/docker/compose/DockerComposeManagerNativeHealthcheckIntegrationTest.java
dansanduleac: 👍

docker-compose-rule-junit4/src/main/java/com/palantir/docker/compose/DockerComposeRule.java
ferozco: I looked into moving parts of `docker-compose-rule-core` into a separate API package and making `docker-compose-rule-junit4` take an implementation dependency on `docker-compose-rule-core` and an api dependency on `docker-compose-rule-api`. It doesn't seem worth while since DCM re-exposes most of `docker-compose-rule-core`

dansanduleac:
👍

PR #304, opened 2019-07-16
Add docker-compose-junit-jupiter package

ferozco:
## Before this PR
We did not support junit-jupiter and forced consumers to use junit-vintage if they wanted to use docker-compose-rule.

Blocked on #303. Once it merges, I'll rebase this PR

## After this PR
==COMMIT_MSG==
Add new `docker-compose-junit-jupiter` package which exposes a junit-jupiter compatible `DockerComposeExtension`
==COMMIT_MSG==

## Possible downsides?
N/A


docker-compose-junit-jupiter/src/integrationTest/java/com/palantir/docker/compose/DockerComposeExtensionIntegrationTest.java
```diff
@@ -0,0 +1,183 @@
...
+import com.palantir.docker.compose.execution.DockerComposeRunArgument;
+import com.palantir.docker.compose.execution.DockerComposeRunOption;
+import java.io.IOException;
+import java.util.List;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+public class DockerComposeExtensionIntegrationTest {
```
ferozco: leave comment describing duplicate classes

docker-compose-junit-jupiter/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+
+testSets {
+    integrationTest
+}
+
+build.dependsOn integrationTest
+
+dependencies {
+    compile project(':docker-compose-rule-core')
+    compile 'org.junit.jupiter:junit-jupiter'
```
dansanduleac: can we just use junit-jupiter-api?
https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter-api

ferozco:
I'll release an RC once this merges

ferozco:
Looks like theres a flake :(

docker-compose-junit-jupiter/src/main/java/com/palantir/docker/compose/DockerComposeExtension.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public void afterEach(ExtensionContext context) throws Exception {
+        after();
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder extends ImmutableDockerComposeExtension.Builder implements BuilderExtensions {
```
CRogers: We should definitely just extend a single builder directly. You can get the builder to return a `DockerComposeManager` subclass like `DockerComposeExtension` by fiddling with one of the immtuables style options.

ferozco: I don't believe that's possible since the immutable classes do not have a common class heirarchy

CRogers: Yeah, it looks that way. My big worry is about all the duplication of common builder helper methods in 3 places - I'm wondering whether there is any way we can not repeat them using interface default methods or something similar.

PR #318, opened 2019-08-07
Added Windows default paths for docker.exe and docker-compose.exe

tifoha:
## Before this PR
this docker-compose-rule not working out of the box with windows machines, cause there are no default path to docker executables

## After this PR
now it should work with windows

## Possible downsides?
didn't find any downsides


ferozco:
@tifoha thanks for the contribution! There was a flake in the build that was resolved by https://github.com/palantir/docker-compose-rule/pull/320, you'll want to rebase your branch to pick up the fix

tifoha:
I merged changes upstream into my fork.
Also I fixed warnings from gradle about using Guava Concurent hash set instead of standard JDK.
@ferozco Can you review my changes?

PR #323, opened 2019-08-14
Bump down conjure java runtime serailization dependency version

CRogers:
## Before this PR
When upgrading people to DCR 1.2.0 internally, they are being forced to take a very recent version of conjure-java-runtime that has compile breaking changes, so making our excavators fail.

## After this PR
==COMMIT_MSG==
Depend on older version of conjure-java-runtime serialization lib to not cause transitive breaks for users of conjure-java-runtime.
==COMMIT_MSG==


CRogers:
:+1:

versions.props
```diff
@@ -13,7 +13,7 @@ one.util:streamex = 0.6.8
 org.assertj:assertj-core = 3.13.2
 com.palantir.conjure:conjure = 4.6.2
 com.palantir.conjure.java:* = 3.15.0
-com.palantir.conjure.java.runtime:* = 4.32.0
+com.palantir.conjure.java.runtime:conjure-java-jackson-serialization = 4.18.1
```
iamdanfox: `# dependency-upgrader:OFF`

PR #331, opened 2019-08-21
Remove duplication by inverting BuilderExtensions control

CRogers:
## Before this PR
Lots of duplication of logic in various builders

## After this PR
By inverting how BuilderExtensions work we can have the individual builders inherit the common behaviour

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/DockerComposeManager.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/DockerComposeManager.java
@@ -281,47 +281,18 @@ public abstract class DockerComposeManager {
         return dockerCompose().run(options, containerName, arguments);
     }
 
-    public interface BuilderExtensions {
-        BuilderExtensions file(String dockerComposeYmlFile);
+    public interface BuilderExtensions<B> {
+        B files(DockerComposeFiles files);
 
-        /**
-         * Save the output of docker logs to files, stored in the <code>path</code> directory.
-         *
-         * See {@link LogDirectory} for some useful utilities, for example:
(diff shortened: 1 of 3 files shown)


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/DockerComposeManager.java
```diff
@@ -281,47 +281,18 @@ public String run(DockerComposeRunOption options, String containerName,
         return dockerCompose().run(options, containerName, arguments);
     }
 
-    public interface BuilderExtensions {
-        BuilderExtensions file(String dockerComposeYmlFile);
+    public interface BuilderExtensions<B> {
```
ferozco: nit: type T instead of B?

CRogers: B for Builder was what I was going for :p

CRogers: TSelf might actually be a better description of what it is

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/DockerComposeManager.java
```diff
@@ -281,47 +281,18 @@ public String run(DockerComposeRunOption options, String containerName,
         return dockerCompose().run(options, containerName, arguments);
     }
 
-    public interface BuilderExtensions {
-        BuilderExtensions file(String dockerComposeYmlFile);
+    public interface BuilderExtensions<TSelf extends BuilderExtensions<TSelf>> {
```
ferozco: nice

PR #332, opened 2019-08-22
junit5 DockerComposeExtension should be a @ClassRule equivalent (also docs)

iamdanfox:
## Before this PR

DockerComposeExtension would try and spinup/tear down containers around every single @Test method.  This is actually a very uncommon use-case internally (like 4 usages or @Rule compared to >700 of @ClassRule).

## After this PR
==COMMIT_MSG==
DockerComposeExtension implements BeforeAllCallback, to ensure containers are started up once at the beginning of the test class
==COMMIT_MSG==

annoyingly the JUnit5 APIs don't give us flexibility to support both choices, so we're stuck with picking one for now

### DIFF
--- a/docker-compose-junit-jupiter/src/integrationTest/java/com/palantir/docker/compose/DockerComposeExtensionIntegrationTest.java
+++ b/docker-compose-junit-jupiter/src/integrationTest/java/com/palantir/docker/compose/DockerComposeExtensionIntegrationTest.java
@@ -16,8 +16,11 @@
 
 package com.palantir.docker.compose;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import com.palantir.docker.compose.configuration.DockerComposeFiles;
 import com.palantir.docker.compose.connection.waiting.HealthChecks;
+import org.junit.jupiter.api.Order;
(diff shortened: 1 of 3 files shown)


docker-compose-junit-jupiter/src/integrationTest/java/com/palantir/docker/compose/DockerComposeExtensionIntegrationTest.java
```diff
@@ -16,169 +16,37 @@
...
     @Test
-    public void run_returns_output() throws Exception {
-        String actual = docker.run(
-                DockerComposeRunOption.options("--entrypoint", "echo"),
-                CONTAINERS.get(0),
-                DockerComposeRunArgument.arguments("hello"));
-        assertThat(actual).isEqualTo("hello");
+    @Order(2)
+    public void container_stays_up_between_tests() {
+        assertThat(docker.containers().container("db").port(5432).getExternalPort()).isEqualTo(port);
```
CRogers: docker tries to reuse ports when they become free, so if you really wanted to be a correct test it probably makes more sense to use container id instead

PR #384, opened 2019-10-22
always copy logs to circle in finally

pnepywoda:
doesnt work - jupiter always calls after anyway


pnepywoda:
Running `DockerComposeExtensionIntegrationTestWithFailingCommand` results in:
```
[main] ERROR com.palantir.docker.compose.reporting.RunRecorder - EnhancedDockerComposeRule has failed in after()
com.palantir.logsafe.exceptions.SafeIllegalArgumentException: Some required fields have not been set: {missingFields=[startTime]}
	at com.palantir.docker.compose.report.DockerComposeRun.validateFields(DockerComposeRun.java:174)
	at com.palantir.docker.compose.report.DockerComposeRun.<init>(DockerComposeRun.java:49)
	at com.palantir.docker.compose.report.DockerComposeRun.<init>(DockerComposeRun.java:24)
	at com.palantir.docker.compose.report.DockerComposeRun$Builder.build(DockerComposeRun.java:306)
	at com.palantir.docker.compose.reporting.RunRecorder.after(RunRecorder.java:73)
	at com.palantir.docker.compose.DockerComposeManager.after(DockerComposeManager.java:269)
	at com.palantir.docker.compose.DockerComposeExtension.afterAll(DockerComposeExtension.java:45)
```

because we reset the `runRecorder` in [after](https://github.com/palantir/docker-compose-rule/blob/master/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/reporting/RunRecorder.java#L70)

CRogers:
We should make sure calling `after()` multiple times is ok. I think the easiest way to do this to either record a `boolean` flag for whether `after` has been called before (and then not run it again) or use a memoised supplier to just run the code in `after` once.

docker-compose-junit-jupiter/src/main/java/com/palantir/docker/compose/DockerComposeExtension.java
```diff
@@ -32,7 +32,12 @@
 
     @Override
     public void beforeAll(ExtensionContext unused) throws IOException, InterruptedException {
-        before();
+        try {
+            before();
+        } catch (RuntimeException e) {
+            after();
+            throw e;
+        }
```
CRogers: I would avoid having behaviour like this in junit5 version only. Instead this should live in `DockerComposeManager.before()` so both junit4 and 5 have the same behaviour

pnepywoda:
@CRogers I can do the call-after-once fix you suggested but do you know why the current integration test isn't failing here? I guess it's not running my new DockerComposeExtensionIntegrationTestWithFailingCommand ?

docker-compose-junit-jupiter/src/main/java/com/palantir/docker/compose/DockerComposeExtension.java
```diff
@@ -30,14 +30,24 @@
 public abstract class DockerComposeExtension extends DockerComposeManager
         implements BeforeAllCallback, AfterAllCallback {
 
+    private boolean hasCalledAfterMethod;
```
CRogers: I think this should live in `DockerComposeManager`, not `DockerComposeExtension` that way `DockerComposeRule` gets this behaviour too.

PR #394, opened 2019-11-14
split compile into api/implementation, stop depending on conjure

pnepywoda:


docker-compose-rule-junit4/build.gradle
```diff
@@ -7,10 +7,13 @@ testSets {
...
-    compile project(':docker-compose-rule-core')
-    compile 'junit:junit'
+    api project(':docker-compose-rule-events-api:docker-compose-rule-events-api-objects')
+    api project(':docker-compose-rule-core')
 
-    testCompile 'org.hamcrest:hamcrest-core'
-    testCompile 'org.mockito:mockito-core'
-    testCompile 'org.assertj:assertj-core'
+    implementation 'com.google.guava:guava'
+    implementation 'junit:junit'
```
pkoenig10: I think this needs to be an API dependency because `DockerComposeRule` is a JUnit rule.

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/reporting/ReportPoster.java
```diff
@@ -18,11 +18,10 @@
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.palantir.conjure.java.serialization.ObjectMappers;
 import com.palantir.docker.compose.report.Report;
 
 class ReportPoster {
-    private static final ObjectMapper OBJECT_MAPPER = ObjectMappers.newClientObjectMapper();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
CRogers: Does this definitely produce the same json output?

CRogers:
Curious: is there any particular reason that the `conjure-java-jackson-serialization` dependency caused problems? It only depends on jackson and safe-logging:preconditions.

pnepywoda:
removing conjure isn't strictly necessary, it's just tech-debt that you hit in https://github.com/palantir/docker-compose-rule/pull/323 (yeah it's a small lib but having the fewest external dependencies makes for an easier time dealing with jar conflicts)

Issue #411, opened 2019-12-04
Port matching is breaks when docker-compose ps output changes


hirochri:
## What happened?

I was using the docker compose rule and found that after adding health checks to my `docker-compose.yml` the docker compose rule fails because it cannot it can't find the internal port for a random container and then triggers this exception:

https://github.com/palantir/docker-compose-rule/blob/a2adeae9ac12b7fc4c9f7a8d1bacaa7db97a430a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Container.java#L91

Did some digging, and came to the conclusion that this breaks because there is some flawed logic in the parsing of the ports from `docker-compose` output.

1. We grab the process output from running `docker-compose ps` here and pass it into the `parseFromDockerComposePs` method: https://github.com/palantir/docker-compose-rule/blob/446e3632ed2ecb9403dcf7f74c325888ab1273ef/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java#L228

2. Within `parseFromDockerComposePs`, a regex is used to match the ports within the output
https://github.com/palantir/docker-compose-rule/blob/f07795c7007f0b7da49cae566cb12fba12f779d2/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Ports.java#L30

When there are no health checks within the `docker-compose.yml`, the output often looks like this:

```
        Name                    Command           State           Ports         
--------------------------------------------------------------------------------
container.name.here   service/bin/linux-        Up      0.0.0.0:1234->1234/tcp
                        amd64/go ...                            
```

Here the regex matches successfully and pulls port `1234` out of `0.0.0.0:1234->1234/tcp`

When there are health checks, the output for `docker-compose ps` gets shifted and truncated at a random spot, causing the port information to be split across multiple lines:

```
        Name                 Command             State              Ports       
--------------------------------------------------------------------------------
container.name.here   service/bin/linux-     Up (healthy)   0.0.0.0:1234->1234/
                         amd64/go ...                       tcp
```

This makes it so that the regex is unable to be matched, so no ports are returned from `parseFromDockerComposePs`, and then the entire docker compose rule fails (as mentioned above).

Here is a test case that confirms this is the issue:
```
 @Test
    public void testcase() {
        Pattern PORT_PATTERN = Pattern.compile("((\\d+).(\\d+).(\\d+).(\\d+)):(\\d+)->(\\d+)/tcp");

[...]

PR #414, opened 2019-12-06
Improve container port and name parsing

hirochri:
## Before this PR
Context: https://github.com/palantir/docker-compose-rule/issues/411
TLDR: The `docker-compose-rule` heavily depends on command output that can be affected by whitespace and thus result in broken parsing..

## After this PR
==COMMIT_MSG==
Fix container port and name parsing by using more effective docker commands and grabbing only the information we care about.
==COMMIT_MSG==


hirochri:
Split into multiple PRs

PR #418, opened 2019-12-09
Improve ports parsing

hirochri:
## Before this PR
This PR fixes: https://github.com/palantir/docker-compose-rule/issues/411
TLDR: Getting port information for various services depends heavily on well white-spaced output from `docker-compose ps <service_name>` and regex, however this quickly breaks when the `ps` command output gets truncated as more information gets added to the `ps` output (health checks, etc).

## After this PR
Instead of applying a regex to unreliable output from `docker-compose ps <service_name>` to get port information, we now:
1. Get the container ID for the service by calling 
`docker-compose ps -q <service_name>`
2. Grab *only* the port information by calling 
`docker ps --no-trunc --format {{.Ports}} --filter id=<container_id>`
3. Apply the regex to the port information string -> get ports

This approach uses docker commands to do more heavily lifting and only return the specific information (ports) that we care about

==COMMIT_MSG==
Improve port parsing by leveraging more specific docker commands.
==COMMIT_MSG==

## Possible downsides?
There is technically a dev break because I modified one of the `DefaultDockerCompose` constructors to take in a `DockerExecutable`.


hirochri:
This one is ready for review @CRogers!

We need to re-trigger changelog-bot somehow, it should be `improvement` and `break` with the message `Improve port parsing by leveraging more specific docker commands.`

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Ports.java
```diff
@@ -48,9 +46,8 @@ public Ports(DockerPort port) {
         return ports.stream();
     }
 
-    public static Ports parseFromDockerComposePs(String psOutput, String dockerMachineIp) {
-        Preconditions.checkArgument(!Strings.isNullOrEmpty(psOutput), "No container found");
-        Matcher matcher = PORT_PATTERN.matcher(psOutput);
+    public static Ports parseFromPortInformation(String portInformation, String dockerMachineIp) {
```
CRogers: Unfortunately nearly everything in DCR is public rather than package-private - so we should do some quick github searches to find out if anyone is using this internally or externally before making such a breaking change.

CRogers: Fortunately in this case it does look safe, but worth bearing in mind!

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -244,9 +263,20 @@ private static boolean downCommandWasPresent(String output) {
         return !output.contains("No such command");
     }
 
-    private String psOutput(String service) throws IOException, InterruptedException {
-        String psOutput = command.execute(Command.throwingOnError(), "ps", service);
-        validState(!Strings.isNullOrEmpty(psOutput), "No container with name '" + service + "' found");
-        return psOutput;
+    private static ErrorHandler swallowingServiceDoesNotExist() {
+        return (exitCode, output, commandName, commands) -> {
+            if (exitCode == 1 && output.isEmpty()) {
```
CRogers: Why exit code 1 and not all exit codes? Will this silently fail if another exit code is thrown?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -244,9 +263,20 @@ private static boolean downCommandWasPresent(String output) {
...
+                // Note: This (badly) checks if the command most likely failed due to
+                // the service not existing. If the ErrorHandler had access to
+                // error output, the proper way to check this would be to inspect
+                // the error output for "ERROR: No such service: <service_name>"
+
+                String fullCommand = Joiner.on(" ").join(Lists.asList(commandName, commands));
+                String serviceName = commands[commands.length - 1];
+
+                log.warn("It looks like `{}` failed.", fullCommand);
+                log.warn("This probably happened because no `{}` service exists.", serviceName);
```
CRogers: Shouldn't this throw rather than silently continue? I imagine we can't recover from this?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
```diff
@@ -45,102 +47,111 @@
...
+                .build(),
+                dockerMachine);
     }
 
-    public DefaultDockerCompose(DockerComposeExecutable rawExecutable, DockerMachine dockerMachine) {
-        this.rawExecutable = rawExecutable;
-        this.command = new Command(rawExecutable, log::trace);
+    public DefaultDockerCompose(
+            DockerComposeExecutable dockerComposeRawExecutable,
+            DockerExecutable dockerRawExecutable,
```
CRogers: This also seems safe from a brief github search (I think 1 person has used it externally and none internally)

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
```diff
@@ -166,37 +172,69 @@ public void throw_exception_when_down_fails_for_a_reason_other_than_the_command_
     @Test
     public void use_the_remove_volumes_flag_when_down_exists() throws IOException, InterruptedException {
         compose.down();
-        verify(executor).execute("down", "--volumes");
+        verify(dockerComposeExecutor).execute("down", "--volumes");
     }
 
     @Test
-    public void parse_the_ps_output_on_ports() throws IOException, InterruptedException {
+    public void get_correct_ports() throws IOException, InterruptedException {
```
CRogers: Not sure about the value of this test - it mocks out a lot of behaviour and I wonder if an integeration test against docker or just testing the parsing directly would be more valuable.

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
```diff
@@ -166,37 +172,69 @@ public void throw_exception_when_down_fails_for_a_reason_other_than_the_command_
...
+        when(dockerComposeExecutedProcess.getInputStream()).thenReturn(toInputStream(""));
+        when(dockerComposeExecutedProcess.exitValue()).thenReturn(1);
+
+        exception.expect(IllegalStateException.class);
+        exception.expectMessage("No container ID found for service with name 'service_that_doesnt_exist'.");
+        compose.ports("service_that_doesnt_exist");
+    }
+
+    @Test
+    public void attempt_to_get_ports_but_all_ports_are_open_for_container() throws IOException, InterruptedException {
```
CRogers: Again here there is so much mocking I'm not sure what this is actually testing?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/Ports.java
hirochri: Ah yeah, missed that one, sorry. Can include explicit comments about usage / impact of the break in the future

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
hirochri: Yeah this is pretty specific, its really only trying to add debug information for when docker can't find a service 
```
~/code/foundry-academy-backend
$ docker-compose ps some_service
ERROR: No such service: some_service
~/code/foundry-academy-backend
$ echo $?
1
```

Seems it would be more proper for this function to check for various failure exit codes and throw an exception explicitly, while still providing output with suggestions around what the potential failure mode was?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
hirochri: If that output is produced, the command will have failed with empty output and then will trigger an exception right after it here: https://github.com/palantir/docker-compose-rule/pull/418/files#diff-cf7a91b7706a65a0c4cc650e69cf5d81R242

I agree that maybe a little misleading and we could clarify how `swallowingServiceDoesNotExist` handles the case

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DefaultDockerCompose.java
hirochri: Yeah it seems reasonable enough, will add comments about break-impact next time!

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
hirochri: I could trade this one out for an integration test, it would be easy enough to create a `docker-compose.yml` with various port setups and then check all the parsing without mocking. Will add that

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/DockerComposeShould.java
hirochri: Will move to an integration test, but this one checks if you specified no ports for your container. An integration test should make this more clear

hirochri:
Swapped out the ports unit tests for integration tests. Not sure why the `DockerComposeManagerNativeHealthcheckIntegrationTest` integration test flaked this time.. Will address if it continues to fail, could you potentially trigger a re-run please?

Otherwise, will fix the error handling mentioned above with whatever we agree upon

CRogers:
While I think this is nicer from a code point of view, it does have a couple of breaks (including using `docker` as well as `docker-compose` for the default code path, which _should_ be fine but you never know), and since #424 has fixed the same problem I'm going to close it out for now.

PR #421, opened 2019-12-12
Run windows tests on CircleCI instead of AppVeyor 

iamdanfox:


iamdanfox:
Seems like we need to do something to get docker to work inside these CircleCI windows workers.

PR #424, opened 2019-12-16
Fix parsing of docker-compose output.

GrahamDennis:
## Before this PR
Newer versions of docker-compose (at least 1.25.0-rc4) adjust the output based on the number of columns available in the terminal. This interferes with parsing of the output and can cause errors like "No internal port X for container Y" when the port does exist.  This is because of a parse failure while parsing the output of `docker-compose ps`.

## After this PR
==COMMIT_MSG==
Set the `COLUMNS` environment variable to an artificially large value for the `docker-compose` process.  This works around newer versions of docker-compose (at least 1.25.0-rc4+) which adjust their output based on the width of the terminal.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-424.v2.yml
+++ b/changelog/@unreleased/pr-424.v2.yml
@@ -1,0 +1,9 @@
+type: fix
+fix:
+  description: |+
+    Fix parsing of docker-compose output
+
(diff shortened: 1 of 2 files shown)


GrahamDennis:
@CRogers: thanks! Note that the past few releases aren't getting published by Circle. It seems like the builds are getting stuck. See https://circleci.com/gh/palantir/docker-compose-rule/5316

PR #495, opened 2020-05-06
Add option to omit the project name

fawind:
## Before this PR

If not set, the `ProjectName` uses `ProjectName#random` to set the docker-compose project name.

In some cases, you want to omit the project name entirely, which makes docker-compose use the name of its current directory as the project name ([reference](https://docs.docker.com/compose/reference/overview/#use--p-to-specify-a-project-name)).

## After this PR

Allow omitting the project name by using `ProjectName#omit`:

```java
DockerComposeExtension.Builder builder = DockerComposeExtension.builder()
        .file("../docker-compose.yml")
        .projectName(ProjectName.omit())
        .build();
```

In this case docker-compose will be run without the `--project-name` option. 

==COMMIT_MSG==
Add option to omit the project name
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-495.v2.yml
+++ b/changelog/@unreleased/pr-495.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add option to omit the project name
+  links:
+  - https://github.com/palantir/docker-compose-rule/pull/495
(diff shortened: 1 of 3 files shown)


CRogers:
:+1:, but I would just check that this solves the problem on the internal repo before merging.

PR #511, opened 2020-07-14
Also add fixed COLUMNS to localMachine

cfredri4:
To fix issue with docker-compose when system terminal is narrow. Also harmonize code according to RemoteBuilder.build.

## Before this PR
Parsing of docker-compose output fails when system terminal is narrow

## After this PR

==COMMIT_MSG==
Make parsing of docker-compose output independent of local system terminal column width.
==COMMIT_MSG==

## Possible downsides?
Probably none.

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
@@ -100,12 +100,19 @@ public class DockerMachine implements DockerConfiguration {
         public DockerMachine build() {
             dockerType.validateEnvironmentVariables(systemEnvironment);
             AdditionalEnvironmentValidator.validate(additionalEnvironment);
-            Map<String, String> combinedEnvironment = newHashMap();
-            combinedEnvironment.putAll(systemEnvironment);
-            combinedEnvironment.putAll(additionalEnvironment);
 
             String dockerHost = systemEnvironment.getOrDefault(DOCKER_HOST, "");
-            return new DockerMachine(dockerType.resolveIp(dockerHost), ImmutableMap.copyOf(combinedEnvironment));
+            String hostIp = dockerType.resolveIp(dockerHost);
+
+            Map<String, String> environment = ImmutableMap.<String, String>builder()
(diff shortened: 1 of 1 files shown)


cfredri4:
Unit tests are failing but this seems unrelated?

docker-compose-rule-core/src/main/java/com/palantir/docker/compose/connection/DockerMachine.java
```diff
@@ -100,12 +100,19 @@ public LocalBuilder withEnvironment(Map<String, String> newEnvironment) {
...
             AdditionalEnvironmentValidator.validate(additionalEnvironment);
-            Map<String, String> combinedEnvironment = newHashMap();
-            combinedEnvironment.putAll(systemEnvironment);
-            combinedEnvironment.putAll(additionalEnvironment);
 
             String dockerHost = systemEnvironment.getOrDefault(DOCKER_HOST, "");
-            return new DockerMachine(dockerType.resolveIp(dockerHost), ImmutableMap.copyOf(combinedEnvironment));
+            String hostIp = dockerType.resolveIp(dockerHost);
+
+            Map<String, String> environment = ImmutableMap.<String, String>builder()
```
fawind, line 107: Thanks for the PR @cfredri4!

For reference, this is the failing test:
https://github.com/palantir/docker-compose-rule/blob/915c47f79a271a3674f243ed5fd923f1bf153bea/docker-compose-rule-core/src/test/java/com/palantir/docker/compose/connection/LocalBuilderShould.java#L114-L128

It fails because it expects to override variables in the `systemEnvironment` with variables from the `additionalEnvironment` when both contain the same key. Previously, we used a `HashMap` for combining the environments, which allowed to overwrite an existing key with a new key-value pair.
When using `ImmutableMap#builder`, this is not allowed and results in an exception.

I would continue using the `HashMap` for building the combined environment as before.

cfredri4: Got it! Did this change in order to harmonize LocalBuilder and RemoteBuilder.
Now changed to allow same functionality in RemoteBuilder instead. ;-)

PR #741, opened 2024-02-09
Remove project name validation

pkoenig10:
The project name validation no longer matches the Docker Compose behaivor.

https://github.com/docker/compose/pull/5788

Docker Compose will normalize project names if they do not conform to it's requirements.

This is better for consumers because it means they can use the same project names in tests and when running Docker Compose manually.


pkoenig10:
I know this project is old. But we still use it in some legacy tests internally and relaxing this validation would be help us out.

PR #744, opened 2024-02-09
Use docker compose v2 by default everywhere

pkoenig10:
Missed this in https://github.com/palantir/docker-compose-rule/pull/729

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerComposeExecutable.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerComposeExecutable.java
@@ -108,7 +108,7 @@ public abstract class DockerComposeExecutable implements Executable {
 
     @Value.Default
     public boolean useDockerComposeV2() {
-        return false;
+        return true;
     }
 
     @Override


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/DockerComposeExecutable.java
```diff
@@ -108,7 +108,7 @@ protected List<String> dockerComposePath() {
 
     @Value.Default
     public boolean useDockerComposeV2() {
-        return false;
+        return true;
```
CRogers, line 111: Definitely need changelog for a change like this.

pkoenig10: We already did this intentionally in https://github.com/palantir/docker-compose-rule/pull/729. We just forgot to change the default here.

CRogers: Eh, I still think it's useful for every human authored change to have changelog, it really helps when builds break and you can quickly see what happened.

CRogers:
:+!:

PR #751, opened 2026-02-09
[Draft] Add stderr to process result output

mdaudali:
## Before this PR

We don't have access to the stderr. When the process failed internally, the stdout made it seem like the error was related to something else (stdout!), which was misleading.
 
## After this PR
Output returns stderr

==COMMIT_MSG==
==COMMIT_MSG==

## Possible downsides?
It's a draft to discuss the format of the output:
1. Changing it in this way now breaks anyone relying exactly on only getting the stdout, in both the output and the errorHandler
2. I've marked empty as <empty> so it doesn't appear like there was an stderr that just got cutoff, but could equally just omit it entirely if no stderr. Chose this way because then it's explicit there's no stderr, rather than unsure if it just got missed e.t.c or a wrong plugin version. 
3. Little bit more verbose now than before.
4. Includes it on a zero status code, though it's valid for programs to output to stderr on zero and it can be potentially useful (or alternatively framed, misleading to hide it)

I'll update the remainder of the tests once agreed on the format!

### DIFF
--- a/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/Command.java
+++ b/docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/Command.java
@@ -80,10 +80,20 @@ public final class Command {
     }
 
     private String processOutputFrom(Process process) {
-        return asReader(process.getInputStream())
+        String stdout = asReader(process.getInputStream())
                 .lines()
                 .peek(logConsumer)
                 .collect(Collectors.joining(System.lineSeparator()));
+
+        String stderr = asReader(process.getErrorStream())
+                .lines()
+                .peek(logConsumer)
+                .collect(Collectors.joining(System.lineSeparator()));
+
+        return String.format(
+                "STDERR: %s\n\n------\n\nSTDOUT: %s\n",
+                stderr.isEmpty() ? "<empty>" : stderr,
+                stdout.isEmpty() ? "<empty>" : stdout);
     }
 
     private static String waitForResultFrom(Future<String> outputProcessing) {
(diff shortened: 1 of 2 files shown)


docker-compose-rule-core/src/main/java/com/palantir/docker/compose/execution/Command.java
```diff
@@ -80,10 +80,20 @@ private ProcessResult run(String... commands) throws IOException, InterruptedExc
...
                 .peek(logConsumer)
                 .collect(Collectors.joining(System.lineSeparator()));
+
+        String stderr = asReader(process.getErrorStream())
+                .lines()
+                .peek(logConsumer)
+                .collect(Collectors.joining(System.lineSeparator()));
+
+        return String.format(
+                "STDERR: %s\n\n------\n\nSTDOUT: %s\n",
```
felixdesouza, line 94: can we just use multi line strings here?

docker-compose-rule-core/src/test/java/com/palantir/docker/compose/execution/CommandShould.java
```diff
@@ -72,7 +73,7 @@ public void return_output_when_exit_code_of_the_executed_process_is_non_0()
         givenTheUnderlyingProcessHasOutput(expectedOutput);
         String commandOutput = dockerComposeCommand.execute(errorHandler, "rm", "-f");
 
-        assertThat(commandOutput).isEqualTo(expectedOutput);
+        assertThat(commandOutput).isEqualTo("STDERR: <empty>\n\n------\n\nSTDOUT: " + expectedOutput + "\n");
```
felixdesouza, line 76: with multi line strings you should be able to do something like:

```java
assertThat(commandOutput).isEqualTo(
  """
  STDERR: <empty>
  
  ------
  
  STDOUT:  %s
  """, expectedOutput);
```

assertj has some nice utility methods to do string formatting inside

felixdesouza: It also gives me a way to see what the actual error message would look like.

So much so, I would actually just put the content for each stream on a new line

e.g.

```
STDERR:
<empty>

------
STDOUT:
%s
```

felixdesouza:
approach seems fine, you *may* end up breaking people if they expect only stdout, so maybe add an option to say "merge output streams" or something

